PR #10, opened 2020-11-02
[WIP] Proposal: a basic implementation of creating control plane components

### LINKED ISSUE #7: Proposal for ControlPlane Custom Resources
This should cover how the all the `Nested*` types work together.

charleszheng44:
This proposal presents a basic/native way of creating components for NestedControlPlane. This is not the final version of the design but can be a good starting point to discuss the details of the implementation. resolve #7


proposals/20201026-creating-control-plane.md
```diff
@@ -0,0 +1,322 @@
...
+
+### Security Model
+
+Creating an NCP requires the end-user to submit a creation request, and the 
+cluster administrator will be responsible for creating the NCP CRs and applying 
+them. Once the NCP is ready, the cluster administrator will return a kubeconfig 
+to the end-user, as each end-user can only access the apiserver assigned to he/she. 
+There is no need to worry about malicious users to manipulate other users' resources. 
+A malicious user can still skew the system by creating a massive amount of resources. 
+To avoid this, we need to enhance the syncer component; however, this topic is
```
weiling61, line 306: Or adding resource quota limit for each user...

proposals/20201026-creating-control-plane.md
```diff
@@ -0,0 +1,322 @@
...
+created based on the NCPT referred by the NCP.
+
+After creating the above prerequisites, sub-controllers will start creating the workload of each component.
+
+### Creation
+
+By default, we will use the StatefulSet to create each component. The readiness 
+and liveness probe will be used, and we will mark each component as ready only 
+when the corresponding Statefuleset is ready. As the KAS cannot work without 
+available Etcd and the KCM cannot run without KAS, the three components need to
```
christopherhein, line 114: If they are created all at once and have liveness/readiness could they be eventually available?

proposals/20201026-creating-control-plane.md
```diff
@@ -0,0 +1,322 @@
...
+create a set of templates in this repository. In the future, we plan to support 
+using templates from some well-known projects, e.g., kubeadm.
+
+
+The creating process will include six steps:
+![Control Plane Creating Process](creating-control-plane.png)
+
+1. The NCP controller notices the creation of a new NCP CR.
+2. NCP controller creates ClusterNamespace, NPKI, NEtcd, NKAS, NKCM, and the NKAS service.
+3. NPKI controller generates TLS certificates for NEtcd and NKAS, creates two kubeconfigs,
```
christopherhein, line 129: I wonder if we could make every component do their own PKI and use these are fields on the components so we could pass them between CRs via reference

christopherhein: we could make the kubeconfig work which should be similar to what KCP does and make that a part of the NCP controller like https://cluster-api.sigs.k8s.io/tasks/certs/using-custom-certificates.html & https://cluster-api.sigs.k8s.io/tasks/certs/generate-kubeconfig.html

this would allow implementers a way to override them

christopherhein: Wild idea… what if NestedEtcd, NestedAPIServer and, NestedControllerManager could end up going the route of being implemented like ClusterAddons where you could have the standard implementation but you could supply patches to them https://github.com/kubernetes-sigs/cluster-addons/blob/master/walkthrough.md flipping the idea away from templates to more declarative and easier to customize. This would mean we could implement higher level abstractions like we have with templates later but staying very focused on the Resources themselves.

munnerz (commented):
I've added a couple of comments 😄 I'm not too sure what the intention of these CRDs are though, beyond being a common aggregation point/pointers to different types of resources. It'd be great to hear what you think here, and whether a model that presents a higher-level abstraction to users would provide more value in the long-run, and allow the project to account for more complex operations (such as upgrades, backups etc).

If the API is too 'low level', implementing something like a backup process could become really difficult and unreliable, so I think it's important it's thought about now 😄

It'd also be good (and perhaps better suited to a larger over-arching doc) to discuss how things like divergences from upstream Kubernetes conformance will be handled in the long run. But I'll bring that to the meeting later perhaps 😄

charleszheng44:
will redesign the CR and submit a new one later

Issue #19, opened 2020-12-01
Create base scaffolding code for all three components


christopherhein:
This should use the design https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/10

PR #24, opened 2020-12-04
Add scaffolding

charleszheng44:
This PR adds the scaffolding generated by `kubebuilder` v3.


christopherhein:
Fixes #19

christopherhein:
We're going to need to regenerate this unfortunately and maybe to make this a little easier we should start with this PR only adding in the one type first, maybe NCP as a blank resource then we can add the rest of the types in individual PRs.

Specifically cause these we're generated with the wrong proper boilerplate so the license headers are messed up - https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/master/hack/boilerplate.go.txt and also the wrong domain and groups. https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/master/PROJECT#L2 should have been `cluster.x-k8s.io` and the groups for NCP, NKAS, NCM, NEtcd should all be under the group `controlplane` https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/24/files#diff-d1fb402f1269ba7396dca9cbd03b44ac055874669bce1d0f239fa2f54ea3584bR6.

christopherhein:
We also should have these kinds under the `v1alpha4` api instead of `v1alpha1`.

api/v1alpha1/nestedapiserver_types.go
```diff
@@ -0,0 +1,70 @@
...
+
+// NestedAPIServerStatus defines the observed state of NestedAPIServer
+type NestedAPIServerStatus struct {
+	// Ready is set if all resources have been created
+	// +kubebuilder:default=false
+	Ready bool `json:"ready,omitempty"`
+
+	// APIServerService is the reference to the service that expose the APIServer
+	// +optional
+	APIServerService *corev1.ObjectReference `json:"apiserverService,omitempty"`
```
christopherhein, line 41: What do you think about not referencing the actual service but changing the `EctdAddress` type to be a shared `Address` type and we use this feed to represent `Endpoint`.

```suggestion
	// Endpoint is the reference to the address that exposes the APIServer
	// +optional
	Endpoint Address `json:"endpoint,omitempty"`
```

api/v1alpha1/nestedetcd_types.go
```diff
@@ -0,0 +1,81 @@
...
+	NestedComponentSpec `json:",inline"`
+}
+
+// NestedEtcdStatus defines the observed state of NestedEtcd
+type NestedEtcdStatus struct {
+	// Ready is set if all resources have been created
+	Ready bool `json:"ready,omitempty"`
+
+	// EtcdDomain defines how to address the etcd instance
+	Addresses []NestedEtcdAddress `json:"addresses,omitempty"`
```
christopherhein, line 38:
```suggestion
	// Addresses defines how to address the etcd instance
	Addresses []Address `json:"addresses,omitempty"`
```

api/v1alpha1/nestedetcd_types.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// CommonStatus allows addons status monitoring
+	addonv1alpha1.CommonStatus `json:",inline"`
+}
+
+// EtcdAddress defines the observed addresses for etcd
+type NestedEtcdAddress struct {
```
christopherhein, line 45:
```suggestion
// Address defines the observed address for this component
type Address struct {
```

api/v1alpha1/nestedapiserver_types.go
charleszheng44: Yes. Sounds good to me

PR #25, opened 2020-12-17
Switch to proper x-k8s.io domain

christopherhein:
Kubernetes won't allow any CRDs to be registered under the `k8s.io` domain, the original this was supposed to be `x-k8s.io` to match other k8s extensions.


vincepri (approved):
/lgtm

PR #29, opened 2021-01-26
NestedControlPlane & NestedCluster Proposal

christopherhein:
Adding new proposal fromm hackmd with regards to how the NCP and NC types are made. This probably still needs a bit of work related to diagrams of interactions but for the most part these types aren't doing the actual orchestration of the control plane.

Closes #7 

Signed-off-by: Chris Hein <me@chrishein.com>

PR #30, opened 2021-01-30
Implement the NestedEtcd controller

charleszheng44:
This PR implements the NestedEtcd controller.


christopherhein (commented):
This looks really good so far, nice work @charleszheng44. Do you want to talk about what you've run into today on the CAPN call?

charleszheng44:
> This looks really good so far, nice work @charleszheng44. Do you want to talk about what you've run into today on the CAPN call?

Yes. We can talk about this PR in today's meeting.

apis/controlplane/v1alpha4/nestedetcd_types.go
```diff
@@ -21,19 +21,23 @@ import (
...
+	// NestedComponentSpec contains the common and user-specified information
+	// that are required for creating the component
 	// +optional
 	NestedComponentSpec `json:",inline"`
 }
 
 // NestedEtcdStatus defines the observed state of NestedEtcd
 type NestedEtcdStatus struct {
-	// Ready is set if all resources have been created
-	Ready bool `json:"ready,omitempty"`
```
christopherhein, line 35: Do we not still want this?

config/rbac/kustomization.yaml
```diff
@@ -6,7 +6,7 @@ resources:
 # Comment the following 4 lines if you want to disable
 # the auth proxy (https://github.com/brancz/kube-rbac-proxy)
 # which protects your /metrics endpoint.
-- auth_proxy_service.yaml
-- auth_proxy_role.yaml
-- auth_proxy_role_binding.yaml
-- auth_proxy_client_clusterrole.yaml
+# - auth_proxy_service.yaml
```
christopherhein, line 9: shouldn't these still be there for the rbac auth proxy?

controllers/controlplane/pki.go
```diff
@@ -0,0 +1,208 @@
+package controlplane
```
christopherhein, line 1: As discussed this will be moved into the NCP controller and leveraging the kubeadm control plane funcs documented in #29

christopherhein (commented):
Couple questions but otherwise looks good.

apis/controlplane/v1alpha4/nestedetcd_types.go
christopherhein: I think I see this was removed cause of the overlap with https://pkg.go.dev/sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/apis/v1alpha1#CommonStatus

charleszheng44: No. I think we can use the `CommonStatus.Phase` instead.

config/rbac/kustomization.yaml
charleszheng44: I will uncomment these four lines.

controllers/controlplane/pki.go
charleszheng44: I will remove this file.

christopherhein (commented):
Looks good only a couple comments, questions.

christopherhein:
We also probably want tests for this.

config/component-templates/nested-etcd/nested-etcd-statefulset-template.yaml
```diff
@@ -0,0 +1,87 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{.nestedetcdName}}
+  namespace: {{.nestedetcdNamespace}}
+spec:
+  replicas: 1
```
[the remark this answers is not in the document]
charleszheng44: Yes. That's a good point

charleszheng44:
> We also probably want tests for this.

test cases added

christopherhein (approved):
Nice, I think we can merge this and keep working on the next few controllers and we can always tweak implementation more as we go.

/approve
/lgtm

PR #37, opened 2021-03-25
Implement the NestedAPIServer controller

charleszheng44:
This PR includes the implementation of the NestedAPIServer controller.


apis/controlplane/v1alpha4/nestedapiserver_types.go
```diff
@@ -0,0 +1,77 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
christopherhein, line 2:
```suggestion
Copyright 2021 The Kubernetes Authors.
```

apis/controlplane/v1alpha4/nestedapiserver_types.go
```diff
@@ -0,0 +1,77 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	addonv1alpha1 "sigs.k8s.io/kubebuilder-declarative-pattern/pkg/patterns/addon/pkg/apis/v1alpha1"
+)
+
+type NestedAPIServerPhase string
+
+const (
+	NestedAPIServerReady   NestedAPIServerPhase = "ready"
+	NestedAPIServerUnready NestedAPIServerPhase = "unready"
+)
```
christopherhein, line 30: Should we reuse the same ones from the etcd controller and make them generic?

apis/controlplane/v1alpha4/nestedapiserver_types.go
```diff
@@ -0,0 +1,77 @@
...
+	// required for creating the component
+	// +optional
+	NestedComponentSpec `json:",inline"`
+}
+
+// NestedAPIServerStatus defines the observed state of NestedAPIServer
+type NestedAPIServerStatus struct {
+	// Ready is set if all resources have been created
+	// +kubebuilder:default=false
+	Ready bool `json:"ready,omitempty"`
```
christopherhein, line 44: Do we still want this given we have `phase`?

apis/controlplane/v1alpha4/nestedapiserver_types.go
```diff
@@ -0,0 +1,77 @@
...
+	// APIServerService is the reference to the service that expose the APIServer
+	// +optional
+	APIServerService *corev1.ObjectReference `json:"apiserverService,omitempty"`
+
+	// CommonStatus allows addons status monitoring
+	addonv1alpha1.CommonStatus `json:",inline"`
+}
+
+//+kubebuilder:object:root=true
+//+kubebuilder:subresource:status
```
christopherhein, line 55: Mapping more towards the nestedetcd types.

```suggestion
//+kubebuilder:object:root=true
//+kubebuilder:resource:scope=Namespaced,path=nestedapiservers,shortName=napiserver
//+kubebuilder:categories=capi,capn
//+kubebuilder:printcolumn:name="Phase",type="string",JSONPath=".status.phase"
//+kubebuilder:printcolumn:name="Age",type="date",JSONPath=".metadata.creationTimestamp"
//+kubebuilder:subresource:status
```

apis/controlplane/v1alpha4/nestedcomponent_types.go
```diff
@@ -0,0 +1,93 @@
...
+	return netcd.Status.CommonStatus
+}
+
+func (netcd *NestedEtcd) IsComponentReady() bool {
+	return netcd.Status.Phase == string(NestedEtcdReady)
+}
+
+func (netcd *NestedEtcd) GetComponentKind() ComponentKind {
+	return Etcd
+}
```
christopherhein, line 56: Maybe we can move these to the specific component types files so that when you are working on a specific type you have the visibility into what might be using it? Unless we can abstract these more and make it only have these defined on the `NestedComponentSpec` since it's embedded.

apis/controlplane/v1alpha4/nestedcontrolplane_types.go
```diff
@@ -28,6 +29,9 @@ type NestedControlPlaneSpec struct {
 	// INSERT ADDITIONAL SPEC FIELDS - desired state of cluster
 	// Important: Run "make" to regenerate code after modifying this file
 
+	// EtcdRef is the reference to the NestedEtcd
+	EtcdRef *corev1.ObjectReference `json:"etcd,omitempty"`
+
```
christopherhein, line 34: Opps, we'll have a couple conflicts between https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/38 and this but we can update them accordingly.

controllers/controlplane/nestedapiserver_controller.go
```diff
@@ -0,0 +1,168 @@
...
+// NestedAPIServerReconciler reconciles a NestedAPIServer object
+type NestedAPIServerReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=nestedapiservers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=nestedapiservers/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=nestedapiservers/finalizers,verbs=update
```
christopherhein, line 44: Hmm, I didn't realize `nestedapiservers/finalizers` was a subresource. til.

controllers/controlplane/nestedapiserver_controller.go
```diff
@@ -0,0 +1,168 @@
...
+//+kubebuilder:rbac:groups=apps,resources=statefulset,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=apps,resources=statefulset/status,verbs=get;update;patch
```
christopherhein, line 46: should we be updating the status?

controllers/controlplane/nestedapiserver_controller.go
```diff
@@ -0,0 +1,168 @@
...
+type NestedAPIServerReconciler struct {
+	client.Client
+	Log    logr.Logger
+	Scheme *runtime.Scheme
+}
+
+//+kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=nestedapiservers,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=nestedapiservers/status,verbs=get;update;patch
+//+kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=nestedapiservers/finalizers,verbs=update
+//+kubebuilder:rbac:groups=apps,resources=statefulset,verbs=get;list;watch;create;update;patch;delete
```
christopherhein, line 45: We're probably missing `services` right?

controllers/controlplane/nestedetcd_controller.go
```diff
@@ -32,14 +27,12 @@ import (
...
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes/scheme"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	ctrlcli "sigs.k8s.io/controller-runtime/pkg/client"
 
-	controlplanev1alpha4 "sigs.k8s.io/cluster-api-provider-nested/apis/controlplane/v1alpha4"
+	capnapis "sigs.k8s.io/cluster-api-provider-nested/apis/controlplane/v1alpha4"
```
christopherhein, line 35: nit. the convention is `clusterv1` eg - https://github.com/kubernetes-sigs/cluster-api/blob/master/controllers/cluster_controller_phases_test.go#L30
```suggestion
	clusterv1 "sigs.k8s.io/cluster-api-provider-nested/apis/controlplane/v1alpha4"
```

christopherhein (commented):
Looks like this is coming together nicely, couple comments from my first pass. We probably need to do some tests still?

apis/controlplane/v1alpha4/nestedapiserver_types.go
charleszheng44: Yes. I think we can let them share the same categories of phases

apis/controlplane/v1alpha4/nestedapiserver_types.go
charleszheng44: Will remove this field from status

apis/controlplane/v1alpha4/nestedcomponent_types.go
charleszheng44: I abstract nested componenets using this interface to simplify the implementation of some common functions. E.g., 

```go
func createNestedComponentSts(..., nc capnapis.NestedComponent, ...) error {}
```.

If we move these to each type, we either need to pass `interface{}` and do type assertion in the function or we need to write `createNestedAPIServerSts` and `createNestedEtcdSts` instead, which might be verbose. Or maybe you think a little verbosity is OK. I may have read too many software engineering blogs, and "don't want to repeat my self" : )

apis/controlplane/v1alpha4/nestedcontrolplane_types.go
charleszheng44: Will modify this field based on #38

controllers/controlplane/nestedapiserver_controller.go
charleszheng44: I didn't realize either. These are generated by Kubebuilder, looks like they make it a subresource

controllers/controlplane/nestedapiserver_controller.go
charleszheng44: We may need to update the CommonStatus.Phase?

controllers/controlplane/nestedapiserver_controller.go
charleszheng44: correct, I will add it

controllers/controlplane/nestedetcd_controller.go
charleszheng44: will change to `clusterv1`

controllers/controlplane/nestedapiserver_controller.go
christopherhein: yeah, agreed

controllers/controlplane/nestedapiserver_controller.go
christopherhein: Wow that's super cool.

apis/controlplane/v1alpha4/nestedcomponent_types.go
christopherhein: Totally, I'm just wondering if because those funcs are exposed on the embedded resource they should already be callable and could be more abstract. So where `IsComponentReady()` has two definitions one for etcd and one for apiserver couldn't that be one with the new reused Phase consts and just check `return obj.Status.Phase == string(NestedComponentReady)`

charleszheng44: aha, I see what you mean. Will remove the `NestedComponent` interface, and move these functions to the embedded struct

christopherhein:
Awesome updates, this looks great. I'm inclined to merge this. Does anyone else want to give it a scan? @Fei-Guo ?

Fei-Guo (approved):
LGTM

PR #38, opened 2021-03-30
Update NCP types

christopherhein:
This is the beginning of the work to define and build the NCP controller, this PR specifically adds the field definitions for the NCP and generates the deepcopy ++ crd yaml.


apis/controlplane/v1alpha4/nestedcontrolplane_types.go
```diff
@@ -17,29 +17,83 @@ limitations under the License.
...
+	// this should contain anything shared between control plane components
+	// +optional
+	APIServer *NestedControlPlaneStatusAPIServer `json:"apiserver,omitempty"`
+
+	// Initialized denotes whether or not the control plane has the
+	// uploaded kubernetes config-map.
+	// +optional
+	Initialized bool `json:"initialized"`
+
+	// Ready denotes that the AWSManagedControlPlane API Server is ready to
```
Fei-Guo, line 61: Update the comment?

apis/controlplane/v1alpha4/nestedcontrolplane_types.go
```diff
@@ -17,29 +17,83 @@ limitations under the License.
...
+	// allows other component controllers to fetch the endpoints.
+	// +optional
+	Etcd *NestedControlPlaneStatusEtcd `json:"etcd,omitempty"`
+
+	// APIServer stores the connection information from the control plane
+	// this should contain anything shared between control plane components
+	// +optional
+	APIServer *NestedControlPlaneStatusAPIServer `json:"apiserver,omitempty"`
+
+	// Initialized denotes whether or not the control plane has the
```
Fei-Guo, line 56: Is the comment valid? What is the config-map?

christopherhein: Good catch, this is a KubeAdm thing where the config is uploaded to a configmap in the kube-system namespace. This might be relevant later but for now you are right and I'll update the comment.

apis/controlplane/v1alpha4/nestedcontrolplane_types.go
christopherhein: doh!

Fei-Guo (approved):
LGTM

Issue #46, opened 2021-05-05
🌱 Setup Prow Presubmits


christopherhein:
We should have our PRs being tested for us with Prow on PR submission so we can have better assurance of the code.

## Links

* Docs: https://github.com/kubernetes/test-infra/blob/master/config/jobs/README.md#adding-or-updating-jobs
* Testing Locally: https://github.com/kubernetes/test-infra/blob/master/prow/cmd/phaino/README.md
* Example: https://github.com/kubernetes/test-infra/blob/master/config/jobs/kubernetes-sigs/cluster-api-provider-aws/cluster-api-provider-aws-presubmits.yaml

Our pre-submit should run `make generate lint test`. This should be called from a script like other Cluster API providers so that we can periodically change what it does without needing to update the `test-infra` repo. EG https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/scripts

PR #48, opened 2021-05-05
🌱 Adding scripts directory for prow jobs

christopherhein:
This add three CI build scripts, there will be a subsequent PR into kubernetes/test-infra to kick off the presubmits.

## Related
Ref #46 


.golangci.yml
```diff
@@ -15,6 +15,12 @@ linters:
     - godot
     - goerr113
     - nestif
+    # TODO(christopherhein) Reenable these and fix errors
```
christopherhein, line 18: Some of these have been cleaned up in #43 so I once both of these are merged we can clean this up.

PR #51, opened 2021-05-06
Do not add slash for REGISTRY with non google cloud

gyliu513:
Fixed https://github.com/kubernetes-sigs/cluster-api-provider-nested/issues/50


christopherhein:
Thanks @gyliu513!

/lgtm
/approve

PR #55, opened 2021-05-14
[VC] unset the VC validation webhook CABundle

charleszheng44:
Leave the CABundle field unset for VC validation webhook configuration, which allows APIServer to use system trusted roots on the APIServer to validate the webhook certificate.


christopherhein (approved):
/lgtm 
/approve

Issue #69, opened 2021-05-22
❓ Investigate v1.20 kubelet Redirection API Deprecation


christopherhein:
Redirection was removed which breaks the vn-agent when we upgrade, we should investigate if our APIs still function as they should or if we need to implement anything "new".

Fei-Guo:
Interesting, what is the new design for streaming api?

christopherhein:
> Interesting, what is the new design for streaming api?

The streaming api still exists, the change was mainly to remove returning a 302 redirect instead to always locally stream through the kubelet as I'm understanding that KEP.

tallclair:
For context, here's the KEP behind the upstream change: https://github.com/kubernetes/enhancements/tree/master/keps/sig-node/1558-streaming-proxy-redirects

PR #70, opened 2021-05-22
⚠️ Update VC controller-runtime 

christopherhein:
This is a relatively large change unfortunately, cause controller runtime's APIs changed a fair amount, adding `context` support throughout. Also as of `v1.20.x` kubelet doesn't support redirects anymore only streaming configs which I have filed https://github.com/kubernetes-sigs/cluster-api-provider-nested/issues/69 to which needs to be followed up on.

## Related
closes #27


virtualcluster/pkg/webhook/virtualcluster/virtualcluster_webhook.go
```diff
@@ -143,6 +144,11 @@ func createValidatingWebhookConfiguration(client client.Client) error {
 	svcPort := int32(constants.VirtualClusterWebhookPort)
 	// reject request if the webhook doesn't work
 	failPolicy := admv1beta1.Fail
+	// use the serviceaccount ca file as the authority
```
Fei-Guo, line 147: This code was removed by #55.  Did you add them back on purpose?

Fei-Guo (commented):
Awesome work.  This change bumps up Kubernetes dependency to 1.20 as well.

virtualcluster/pkg/webhook/virtualcluster/virtualcluster_webhook.go
christopherhein: Oops, thanks for catching that. I don't know how that made it back in.

christopherhein:
> Awesome work. This change bumps up Kubernetes dependency to 1.20 as well.

We still don't support the one annoying thing about the ClusterVersion CR where if you regenerate it using the proper version of `controller-gen` for the version of controller runtime we use then you end up with a huge ClusterVersion CRD and run into issues, so for now in the make file I've disabled generating the CRD manifests so that we continue to use the current working ones.

weiling61 (approved):
LGTM. (Most changes are due to runtime package switch, no major logic modification)

christopherhein:
> LGTM. (Most changes are due to runtime package switch, no major logic modification)

@weiling61 yup, thanks for giving this a read-through. Something to take into consideration is how the new `MCController` is implemented note I've broken the implementation for `List` and `Get` and removed `ListByObjectType` and `GetByObjectType`. Instead using `Get` and `List` which take in a pointer to `client.Object` or `client.ObjectList` respectively which allows us to populate that object instead of having to return the object back and do type assertions. I've also made one other change to remove us having to have a custom `scheme` now all "core" types are supported out of the box without us having https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/70/files#diff-f9cddd7f54205f0c3f3483eb332fb22f9fa9c1184b76ec5223af230b5ab7c14b

Fei-Guo:
“I've also made one other change to remove us having to have a custom scheme now all "core" types are supported out of the box without us having”

Will this make syncing CR resources (not CRD) harder or break existing CR synchronization?

christopherhein:
> Will this make syncing CR resources (not CRD) harder or break existing CR synchronization?

@Fei-Guo this should make it easier to manage these for folks cause we can leverage a lot of the tools that have been built already like controller-runtimes SchemeBuilder which is preconfigured for all kubebuilder/controller runtime projects. This would allow you to take a package like https://github.com/kubernetes-sigs/kubebuilder/tree/master/testdata/project-v2/api/v1 and all you'd need to do is:

```go
import (
  "sigs.k8s.io/kubebuilder/testdata/project-v2/api/v1"
  "sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/scheme"
)

func init() {
  // custom scheme we want to add 
  v1.AddToScheme(scheme.Scheme)
}
```

This will use the SchemeBuilder from https://github.com/kubernetes-sigs/kubebuilder/blob/master/testdata/project-v2/api/v1/groupversion_info.go#L27-L36 and takes in a defined `*runtime.Scheme` to add those new types and GVK for you.

christopherhein:
As I was reviewing that I did make one mistake which I've just cleaned up the `var Scheme = runtime.NewScheme()` was supposed to use https://github.com/kubernetes/client-go/blob/v0.21.1/kubernetes/scheme/register.go#L72 instead which is the client-go scheme that has everything defined already. So I've just updated that. Just waiting on tests to build.

zhuangqh:
thanks for your job! controller runtime style client makes everything simple 😄

zhuangqh (approved):
LGTM

christopherhein:
Awesome, thanks @zhuangqh. Looks like we just need someone to `/lgtm` and we'll get this merged.

PR #72, opened 2021-05-22
Setting kind name as CAPN

gyliu513:
Sometimes people want to run multiple kind clusters to have different test purpose, it is better to add a name for the kind cluster for capn as well.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -19,7 +19,7 @@ cd cluster-api-provider-nested
 ### Create `kind` cluster
 
 ```shell
-kind create cluster
+kind create cluster --name=capn
 ```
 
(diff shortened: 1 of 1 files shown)


Fei-Guo:
Thanks for the fix.

/lgtm

christopherhein:
@gyliu513 looks like you need to rebase. I'm wondering if this is necessary? Since if a person is already running a `kind` they should know this and would have the sense to add the `--name` flag. 🤔  

Either way we should probably add a `Cleanup` section to this doc that does `kind delete cluster --name capn`.

christopherhein:
Thanks for adding this! 🎉 

/lgtm
/approve

PR #83, opened 2021-05-31
syncer: check meta equality for sa type token secret

zhuangqh:


virtualcluster/pkg/syncer/conversion/equality.go
```diff
@@ -517,8 +517,23 @@ func (e vcEquality) CheckBinaryDataEquality(pObj, vObj map[string][]byte) (map[s
...
+			updatedObj.Labels = labels
+		}
+
+		annotations, equal := e.checkDWKVEquality(pObj.Annotations, vObj.Annotations)
+		if !equal {
+			if updatedObj == nil {
+				updatedObj = pObj.DeepCopy()
+			}
+			updatedObj.Annotations = annotations
+		}
```
Fei-Guo, line 536: Line 537, should we return nil or updatedObj?

PR #86, opened 2021-06-03
Install binaries for envTest

Fei-Guo:
The controller-runtime EnvTest requires the Kubernetes binaries to be exist in `/usr/local/kubebuilder/bin`. 
This change adds the support script in Makefile in case they are missing or too old. 

This change also removes the redundant CRDs.


christopherhein (commented):
Awesome job! 🚀 

/lgtm
/approve

PR #87, opened 2021-06-03
🌱  VN Agent Healthz Checks

christopherhein:
This add the ability to use vn-agent with a `kubeconfig` specified, which works for the Deployment style vs the DaemonSet mechanism as well this added a `/healthz` endpoint for better checking of it's availability.

### DIFF
--- a/virtualcluster/cmd/vn-agent/app/server.go
+++ b/virtualcluster/cmd/vn-agent/app/server.go
@@ -30,6 +30,7 @@ import (
 	"k8s.io/component-base/cli/globalflag"
 	"k8s.io/component-base/term"
 	"k8s.io/klog"
+	"k8s.io/kubernetes/pkg/healthz"
 
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/cmd/vn-agent/app/options"
 	utilflag "sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/util/flag"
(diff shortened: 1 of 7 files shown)


virtualcluster/cmd/vn-agent/app/server.go
```diff
@@ -103,6 +104,8 @@ func Run(c *config.Config, serverOption *options.ServerOption, stopCh <-chan str
 		},
 	}
 
+	fmt.Printf(serverOption.ClientCAFile)
```
charleszheng44, line 107: Why we wanna print out the client CA file?

charleszheng44 (changes requested):

virtualcluster/cmd/vn-agent/app/server.go
christopherhein: Doh, debugging lol, I will remove that.

Fei-Guo:
The envTest fails. Looking at the log 


2021-06-03T18:32:39.544Z	INFO	controllers.VirtualCluster.Native	service already exist	{"service": "apiserver-svc"}
STEP: Faking apiserver STS Status Updates
STEP: Creating Control Plane controller-manager StatefulSet
2021-06-03T18:32:45.550Z	INFO	controllers.VirtualCluster.Native	creating secret	{"name": "root-ca", "namespace": "default-137bf3-virtualcluster-samplevxjfk"}


There is a 6 seconds gap between service log and creating secret log which looks weird. I only see one Get call in between from code which can be slow.  

We probably need to increase the timeout in envTest.

Issue #91, opened 2021-06-08
🐛 Post Submit Image Builds


christopherhein:
All Post Submit Image builds are failing because of the wrong GCP project being specified. We need to update the `Makefile` to point to the correct registry URL w/ project name.

PR #93, opened 2021-06-08
🐛  Updating staging and prod registries

christopherhein:
This fixes the builds which were erroring cause of GCP projects for releasing the images.

XRef: #91 


vincepri (approved):
/lgtm

PR #139, opened 2021-06-17
syncer: remove feature guard when checking ns meta equality

zhuangqh:

update namespace meta is a generic operation

Fixes #


virtualcluster/pkg/syncer/resources/namespace/checker.go
```diff
@@ -137,19 +137,18 @@ func (c *controller) PatrollerDo() {
 		// if vc object is deleted, we should reach here
 		if c.shouldBeGarbageCollected(p) || p.Annotations[constants.LabelUID] != string(v.UID) {
 			c.deleteNamespace(p)
```
zhuangqh, line 139: it is unnecessary to check equality for deleting ns

Issue #130, opened 2021-06-15
✨ [VC] Support Releasing Manifests & Tagging for VC code base


christopherhein:
**User Story**

As a user I would like to use blessed manifests for deploying VC for so that I can expect a stable release.

**Detailed Description**

Right now only CAPN is deployed/setup when `make release` is run, this work pairs after #129 to add full release support.

**Anything else you would like to add:**

Removing `ClusterVersion` from VC should be done prior to this work.

/kind feature
/milestone v0.1.x

Issue #129, opened 2021-06-15
✨ Support Staging Releases for VC code base


christopherhein:
**User Story**

As a user I would like to be able to use VC with CAPN from Prow based builds for so that I can use automated and "blessed" images.

**Detailed Description**

Right now CAPN is released every `main` branch merge plus, this is done using the `make release-staging` target, we should update these downstream targets to add CAPN into this.

**Anything else you would like to add:**

This doesn't need to deal with the manifests, this will be handled in the `make release` changes.

/kind feature
/milestone v0.1.x

christopherhein:
/retitle ✨ [VC] Support Staging Releases for VC code base

dongwangdw:
@christopherhein 
Want to confirm with you that: the target here is to:
```
Whenever there is code merge in virtualcluster, the image build should also trigger from CAPN by make release?
```

Because you said: ```this will be handled in the make release changes.```
But what if there is no release at the time when the code changes happened in virutalcluster?
Should we also handle this for staging release?

And btw, I did not find any steps for prod images pushing. 
The release tag just updates the manifests. Did I miss anything?

christopherhein:
The target of this issue is to handle the staging releases similar to what we have in `capn/Makefile` under the target `make staging-release` but integrated to work against VC code base, this will make it so that every merge into `main` will get a sha based release that we can then verify with. For the prod releases it's actually not automated via prow, what we do is follow the guide https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/developer/releasing.md to make sure we tag the release locally and then rerun the `make release` target to generate the production ready manifests, once those are built and the stage was pushed we manually make a Github release and upload the production manifests that were generated locally.

PR #142, opened 2021-06-19
Correct manifest path since the file name has been changed in previously ccommit

vincent-pli:

Fixes #

### DIFF
--- a/virtualcluster/doc/demo.md
+++ b/virtualcluster/doc/demo.md
@@ -59,7 +59,7 @@ kubectl apply -f https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-p
 To create all VirtualCluster components:
 
 ```bash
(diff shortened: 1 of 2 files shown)


christopherhein:
Whoops, this is probably going to need more updates to work correctly, because the images aren't published anywhere this manifest will fail to be used for any cluster that you don't run `make build-images` and thus don't push those images into the kind or minikube cluster, it also requires CAPN which will require the user to do and follow the `docs/README.md`.

Given this isn't ready how should we handle the updates to this? Should we revert since you should be able to use VC w/o CAPN right now via the old readme, and then once we get #129 and #130 ?

christopherhein:
@Fei-Guo @vincent-pli ^^^^

Fei-Guo:
@christopherhein Yes. I think you overwrote the old all_in_one with all_in_one_capi?  can you add the old all_in_one back and revert the README to the old all_in_one?

christopherhein:
Oh shoot, wow. sorry I missed that, yea we can add that manifest back.

PR #145, opened 2021-06-24
🐛 use self-signed certificate for the virtualcluster webhook server

### LINKED ISSUE #125: 🐛 Webhook caBundle issues for Virtual Cluster
Not dig in too much to the code, but in my env, the `webhook` for `VirtualCluster` not work.
I get this when I try to create `VirtualCluster`:
```
Error from server (InternalError): error when creating "virtualcluster_1_nodeport.yaml": Internal error occurred: failed calling webhook "virtualcluster.validating.webhook": Post "https://virtualcluster-webhook-service.vc-manager.svc:9443/validate-tenancy-x-k8s-io-v1alpha1-virtualcluster?timeout=30s": x509: certificate signed by unknown authority
```

Then i check the `ValidatingWebhookConfiguration` and there is no `caBundle` at all in `virtualcluster-validating-webhook-configuration`.

After I modify the `virtualcluster-validating-webhook-configuration` and set the `caBundle` with the cluster's CA, everything works as exppected.

Seems we do not set the `caBundle`, am i missing something?

charleszheng44:

Use a self-signed certificate for the virtual cluster webhook. See issue #125 for more details.

fixes #125

### DIFF
--- a/virtualcluster/config/setup/all_in_one.yaml
+++ b/virtualcluster/config/setup/all_in_one.yaml
@@ -75,6 +75,7 @@ rules:
   - ""
   resources:
   - services
+  - events
   verbs:
   - get
   - list
@@ -182,6 +183,18 @@ rules:
   resources:
   - mutatingwebhookconfigurations
   - validatingwebhookconfigurations
+  verbs:
+  - get
+  - list
+  - watch
+  - create
+  - update
+  - patch
+  - delete
+- apiGroups:
+  - coordination.k8s.io
+  resources:
+  - leases
   verbs:
   - get
   - list
--- a/virtualcluster/pkg/webhook/virtualcluster/virtualcluster_webhook.go
+++ b/virtualcluster/pkg/webhook/virtualcluster/virtualcluster_webhook.go
@@ -17,35 +17,24 @@ limitations under the License.
 package virtualcluster
 
 import (
+	"bytes"
 	"context"
-	"crypto/ecdsa"
-	"crypto/elliptic"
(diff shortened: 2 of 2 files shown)


charleszheng44:
@vincent-pli The VC should be able to run on Kind after this PR is merged.

@christopherhein 
@Fei-Guo

christopherhein:
@charleszheng44 do you think VC should be managing these resources for the user? or should we be prescribing that the operator use cert-manager or manage their cluster certificates themselves? 

I know cert-manager itself did a lot of cludgy work just to enable this type of seamless setup process for auto-deploying their own certs and it was fairly painful to write and maintain. It feels like we could make this easier with just supplying configs to use cert manager auto-injection and then we could rip out all this setup code and expect that if they aren't using cert-manager that they have an existing flow for certs and thus can just use that path.

christopherhein:
Code looks fine otherwise more just commenting on the overall approach.

charleszheng44:
@christopherhein I prefer using cert-manager to manage all certifications. But do you think it would be an overkill to deploy cert-manager to just manage the certification for the webhook server. Or we can use the cert-manager to manage all certifications for VC control plane components. However, I think we will use the CAPN to replace the VC native mode, and maybe we should import the cert-manager by then? We can use the cert-manager to manage certifications for all components (i.e., CAPN control plane components and VC webhook servers).

christopherhein:
@charleszheng44 mmk, that seems like a reasonable approach instead. 

/lgtm

PR #148, opened 2021-06-29
Correct Kind setting for MultiClusterController

vincent-pli:
The original implements can get the `Kind` when creating `MultiClusterController`, the `GroupVersionKind()` is empty.

### DIFF
--- a/virtualcluster/pkg/util/mccontroller/mccontroller.go
+++ b/virtualcluster/pkg/util/mccontroller/mccontroller.go
@@ -36,6 +36,7 @@ import (
 	clientgocache "k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/cache"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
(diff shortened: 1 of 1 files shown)


Fei-Guo:
/hold

Let me double check if this works for CRDs.

virtualcluster/pkg/util/mccontroller/mccontroller.go
```diff
@@ -114,13 +115,17 @@ type ClusterInterface interface {
 
 // NewMCController creates a new MultiClusterController.
 func NewMCController(objectType client.Object, objectListType client.ObjectList, rc reconciler.DWReconciler, opts ...OptConfig) (*MultiClusterController, error) {
-	kind := objectType.GetObjectKind().GroupVersionKind().Kind
+	kinds, _, err := scheme.Scheme.ObjectKinds(objectType)
```
christopherhein, line 118: I believe you should be using this `scheme.Scheme` global variable from - https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/pkg/syncer/util/scheme/scheme.go#L21 which is what other packages reference, this should allow a non-core registered type like a CRD to work.

vincent-pli: @christopherhein 
I believe you are right, the `scheme` from `controller-runtime` or `client-go` do not include CRD's scheme,
thanks

virtualcluster/pkg/util/mccontroller/mccontroller.go
```diff
@@ -31,6 +31,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/kubernetes/scheme"
```
christopherhein, line 34:
```suggestion
	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/scheme"
```

christopherhein (changes requested):

christopherhein (approved):
/lgtm

@Fei-Guo did you check this out?

PR #151, opened 2021-07-01
Correct the description for crd syncer

vincent-pli:
**What this PR does / why we need it**
It's `labels` to enable the `CRD` syncer not `annotation`.


christopherhein:
Thanks @vincent-pli !

PR #158, opened 2021-07-04
Disable nodelifecycle controller to void NotReady status of virtual node

### LINKED ISSUE #154: The syncer report different pod.status regular
Created `deploy` in tenant cluster and `pod` works as expected, but get message regular(5min) in `syncer`'s log:
```
1 checker.go:261] status of pod default-a32f71-vc-sample-1-default/test-deploy-1-696d8b7c77-clz6m diff in super&tenant master
```

Do some research, seems:
1. The `nodelifecycle` controller of `kube-manager-controller` in tenant cluster can not contact the virtual node, then set the status of node as `NotReady`, then add taint to the node and `eviction manager` try to eviction pod in that virtual node, then set the `condition` of the node as `status: false`

2. Then the `pod syncer` diff the status and find the difference, then raise the message and sync the status of pod.
3. After a while the `node syncer` find the node difference and sync the status of node as `Ready`

and the process occurred regular.

So I try to disable the `nodelifecycle` in `kube-controller-manager` of tenant cluster then everything works as expected.

So could we disable the `nodelifecycle`?

/kind bug

vincent-pli:

Fixes #154

### DIFF
--- a/controlplane/nested/component-templates/nested-controllermanager/nested-controllermanager-statefulset-template.yaml
+++ b/controlplane/nested/component-templates/nested-controllermanager/nested-controllermanager-statefulset-template.yaml
@@ -36,6 +36,7 @@ spec:
         - --use-service-account-credentials=true
         - --experimental-cluster-signing-duration=87600h
         - --node-monitor-grace-period=200s
+        - --controllers=*,-nodelifecycle
         - --v=2
         livenessProbe:
           httpGet:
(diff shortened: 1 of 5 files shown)


christopherhein:
/ok-to-test

Would you also add this same flag to - https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/controlplane/nested/component-templates/nested-controllermanager/nested-controllermanager-statefulset-template.yaml#L38 🙏

vincent-pli:
The file you mentioned is already in the change list, please take a look @christopherhein

christopherhein:
> The file you mentioned is already in the change list, please take a look @christopherhein

Doh, I'm sorry about that, missed the first file. Great job!

/lgtm
/approve

PR #161, opened 2021-07-07
Remove old webhookconfiguration if it alreasy exist before create new one

vincent-pli:
In #145 , we use self-signed certificate for `ValidatingWebhookConfiguration`, that's means we will generate self-signed CA when `vc-manager` startup.
Based on current implements, when creating `ValidatingWebhookConfiguration` and found it already exist we will give up to create a new one and expect to use the older one, that's will cause problem since the `private key` is re-generated by new CA but in old `ValidatingWebhookConfiguration` included the old `CA`, that's will case the webhook raise:
```
root@rentz1:~# kubectl delete VirtualCluster vc-sample-1
Error from server (InternalError): Internal error occurred: failed calling webhook "virtualcluster.validating.webhook": Post "https://virtualcluster-webhook-service.vc-manager.svc:9443/validate-tenancy-x-k8s-io-v1alpha1-virtualcluster?timeout=30s": x509: certificate signed by unknown authority (possibly because of "crypto/rsa: verification error" while trying to verify candidate authority certificate "tenancy.x-k8s.io")
```

In my env, the `vc-manager` has restarted by unknown reason, then I hit the error mentioned upper.

Fixes #

### DIFF
--- a/virtualcluster/pkg/webhook/virtualcluster/virtualcluster_webhook.go
+++ b/virtualcluster/pkg/webhook/virtualcluster/virtualcluster_webhook.go
@@ -172,7 +172,17 @@ func createValidatingWebhookConfiguration(client client.Client, caPEM []byte) er
 		if !apierrors.IsAlreadyExists(err) {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


charleszheng44 (approved):
Good catch!
/lgtm

PR #200, opened 2021-07-30
✨ Watch instead of loop check nested cluster

### LINKED ISSUE #137: ✨ Update CAPI VC Controller to Use a Cluster watch.
**User Story**

As an operator I would like to the CAPI Controller for VC to watch the Cluster resource for updates so I don't need to poll the Cluster resource for changes.

**Detailed Description**

Currently in #136 this used `ctrl.Result{RequeueAfter}` which is polling the cache for updates to the Cluster object, this should be done by triggering a change when the Cluster changes instead making the system more efficient.

Xref: #136 

/kind cleanup
/milestone v0.1.x

jichenjc:

Fixes #137

### DIFF
--- a/controllers/nestedcluster_controller.go
+++ b/controllers/nestedcluster_controller.go
@@ -20,15 +20,25 @@ package controllers
 
 import (
 	"context"
-	"time"
+	"reflect"
 
 	"github.com/go-logr/logr"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
 	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/annotations"
+	"sigs.k8s.io/cluster-api/util/predicates"
 	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/builder"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/controller"
+	"sigs.k8s.io/controller-runtime/pkg/event"
+	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
+	"sigs.k8s.io/controller-runtime/pkg/reconcile"
+	"sigs.k8s.io/controller-runtime/pkg/source"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-nested/api/v1alpha4"
 	controlplanev1 "sigs.k8s.io/cluster-api-provider-nested/controlplane/nested/api/v1alpha4"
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -149,7 +150,7 @@ func main() {
 		Client: mgr.GetClient(),
 		Log:    ctrl.Log.WithName("controllers").WithName("infrastructure").WithName("NestedCluster"),
 		Scheme: mgr.GetScheme(),
-	}).SetupWithManager(mgr); err != nil {
+	}).SetupWithManager(ctx, mgr, concurrency(10)); err != nil {
```
charleszheng44, line 153: Do you think we can make the `concurrency` configurable (through the command-line option) ?

controllers/nestedcluster_controller.go
```diff
@@ -48,10 +59,51 @@ type NestedClusterReconciler struct {
...
+				predicate.Funcs{
+					// Avoid reconciling if the event triggering the reconciliation is related to incremental status updates
+					UpdateFunc: func(e event.UpdateEvent) bool {
+						oldCluster := e.ObjectOld.(*infrav1.NestedCluster).DeepCopy()
+						newCluster := e.ObjectNew.(*infrav1.NestedCluster).DeepCopy()
+						oldCluster.Status = infrav1.NestedClusterStatus{}
+						newCluster.Status = infrav1.NestedClusterStatus{}
+						oldCluster.ObjectMeta.ResourceVersion = ""
+						newCluster.ObjectMeta.ResourceVersion = ""
+						return !reflect.DeepEqual(oldCluster, newCluster)
```
charleszheng44, line 79: I think it is OK for now since there is only one field (`Ready`) in the `NestedCluster.Status`. @christopherhein what's your opinion?

controllers/nestedcluster_controller.go
```diff
@@ -48,10 +59,51 @@ type NestedClusterReconciler struct {
...
+	log := ctrl.LoggerFrom(ctx)
+
 	return ctrl.NewControllerManagedBy(mgr).
-		For(&infrav1.NestedCluster{}).
-		Owns(&controlplanev1.NestedControlPlane{}).
+		WithOptions(options).
+		For(&infrav1.NestedCluster{},
+			builder.WithPredicates(
+				predicate.Funcs{
+					// Avoid reconciling if the event triggering the reconciliation is related to incremental status updates
```
christopherhein, line 71: Were you seeing this happening frequently? I haven't experienced this before but I also could have just missed it.

controllers/nestedcluster_controller.go
christopherhein: I think this is fine, cause we don't really need to rereconcile after we set `Ready`.

main.go
christopherhein: I agree, I think we should have this configurable. We can take the `cluster-concurrency` flag from CAPI - https://github.com/kubernetes-sigs/cluster-api/blob/master/main.go#L125-L126

jichenjc: I added a todo here actually in concurreny implemention function ..
will see how we can add this

controllers/nestedcluster_controller.go
jichenjc: this is copied from CAPO repo code, https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/748
I admit I didn't have such as well..

jichenjc:
with workaround from #201 at least I can create a nested cluster follow the dev guide with the new PR..

PR #167, opened 2021-07-12
✨ Enabled Aggregate API for Virtual Cluster

gyliu513:
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-nested/issues/166

@christopherhein @Fei-Guo 

This PR only covered VirtualCluster but does not cover nested control plane, will fix nested control plane in another PR.

@wangjsty ^^

PR #211, opened 2021-08-19
🌱 add test for "Enabled Aggregate API for Virtual Cluster"

Abirdcfly:
Add test for #167
Fixes #


Abirdcfly:
The other thing is that our [demo](https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/doc/demo.md) does not work because of #167. 
Maybe we need to update the docker image in or update the [demo documentation](https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/doc/demo.md) to prompt users to build the latest image themselves

PR #216, opened 2021-09-14
✨ Generate the manifests by calling the kubeadm

charleszheng44:
Generate the ncp manifests using the `kubeadm` executable


Dockerfile
```diff
@@ -45,14 +45,17 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
...
     --mount=type=cache,target=/root/.local/share/golang \
     CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} go build -ldflags "${LDFLAGS} -extldflags '-static'"  -o manager ${package}
+RUN curl -L https://dl.k8s.io/release/v1.22.0/bin/linux/amd64/kubeadm -o kubeadm && chmod +x kubeadm
 ENTRYPOINT [ "/start.sh", "/workspace/manager" ]
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
+# FROM gcr.io/distroless/static:nonroot
+FROM ubuntu
```
christopherhein, line 54: If we need to switch this can we use the debian from the distroless repo - https://github.com/GoogleContainerTools/distroless#docker

christopherhein (commented):
This is amazing so far, 1 comment for now and I want to pul this down and experiment with it tomorrow. 👍

controlplane/nested/controllers/controller_util.go
```diff
@@ -302,3 +306,367 @@ func genObjRefFromObj(obj ctrlcli.Object) corev1.ObjectReference {
...
+	for name, podTemplate := range templates {
+		pod := corev1.Pod{}
+		if err := yamlToObject([]byte(podTemplate), &pod); err != nil {
+			return nil, err
+		}
+		switch name {
+		case kubeadm.APIServer:
+			ret[kubeadm.APIServer] = completeKASPodSpec(pod, clusterName)
+		case kubeadm.ControllerManager:
+			ret[kubeadm.APIServer] = completeKCMPodSpec(pod, clusterName)
```
jzhoucliqr, line 345:
```suggestion
			ret[kubeadm. ControllerManager] = completeKCMPodSpec(pod, clusterName)
```

controlplane/nested/controllers/controller_util.go
```diff
@@ -302,3 +306,367 @@ func genObjRefFromObj(obj ctrlcli.Object) corev1.ObjectReference {
...
+		case kubeadm.Etcd:
+			ret[kubeadm.APIServer] = completeEtcdPodSpec(pod, clusterName)
```
jzhoucliqr, line 347:
```suggestion
			ret[kubeadm. Etcd] = completeEtcdPodSpec(pod, clusterName)
```

jzhoucliqr:
Thanks @charleszheng44 for the effort!  I'm trying out this PR, however, the etcd statefulset DNS resolve have some issue so it doesn't come up, not sure if you've seen this? 

`{"level":"warn","ts":1633452597.3438876,"caller":"netutil/netutil.go:131","msg":"failed to resolve URL Host; returning","url":"https://cluster-s1-etcd-0.cluster-s1-etcd.default.svc:2380","host":"cluster-s1-etcd-0.cluster-s1-etcd.default.svc:2380","retry-interval":1,"error":"lookup cluster-s1-etcd-0.cluster-s1-etcd.default.svc on 10.96.0.10:53: no such host"}`

charleszheng44:
> Thanks @charleszheng44 for the effort! I'm trying out this PR, however, the etcd statefulset DNS resolve have some issue so it doesn't come up, not sure if you've seen this?
> 
> `{"level":"warn","ts":1633452597.3438876,"caller":"netutil/netutil.go:131","msg":"failed to resolve URL Host; returning","url":"https://cluster-s1-etcd-0.cluster-s1-etcd.default.svc:2380","host":"cluster-s1-etcd-0.cluster-s1-etcd.default.svc:2380","retry-interval":1,"error":"lookup cluster-s1-etcd-0.cluster-s1-etcd.default.svc on 10.96.0.10:53: no such host"}`

@jzhoucliqr Thanks for finding the bug. I am working on it and will let you know once it is fixed 😃

jzhoucliqr:
I think the issue is because in statefulset we use component name as service name, which ends with -nestedetcd, but the actual service object name ends with -etcd. 
https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/216/files#diff-d3d34848bfdf7f6567e793bc968606d1628a0c36a69f141ecb7bf09e90c439c3R164

controlplane/nested/kubeadm/consts.go
```diff
@@ -0,0 +1,21 @@
+package kubeadm
+
+const (
+	KubeadmExecPath          = "/kubeadm"
+	KASManifestsPath         = "/etc/kubernetes/manifests/kube-apiserver.yaml"
```
jzhoucliqr, line 5: By using the same default file locations, could it cause conflict, for eg, when multiple clusters are created at the same time?

charleszheng44:
> I think the issue is because in statefulset we use component name as service name, which ends with -nestedetcd, but the actual service object name ends with -etcd. https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/216/files#diff-d3d34848bfdf7f6567e793bc968606d1628a0c36a69f141ecb7bf09e90c439c3R164

That's right 😅  Let me try it out. Thanks

charleszheng44:
@jzhoucliqr It should work now. Would you try it out?

controlplane/nested/kubeadm/consts.go
charleszheng44: Yes. That could be a problem. `kubeadm` does not support generating static pod manifests in a directory other than `/etc/kubernetes/manifests`. We should figure out a way to solve it.

charleszheng44: @christopherhein @Fei-Guo do you have any suggestions?

christopherhein: Can we span a new pod for each cluster and generate the manifests in the new pod then terminate after they have been created as configmaps?

charleszheng44: > Can we span a new pod for each cluster and generate the manifests in the new pod then terminate after they have been created as configmaps?

@christopherhein @jzhoucliqr 

Yes. That's also the only way I can see. Let's try to merge this PR first and I will create a new issue/PR to fix it?

christopherhein: I think that might work the only idea I could have is tracking what cluster is currently being templated and "lock" other reconciliation from processing until the current cluster bring up finishes.

charleszheng44: Then we need to add a big lock in the reconcile loop or reduce the ncp_controller's concurrency to one to ensure the ncp_controller is not creating more than one nested control plane at a time. Will that affect the performance?

jzhoucliqr:
> @jzhoucliqr It should work now. Would you try it out?

Thanks. Will try it out.

Issue #215, opened 2021-09-13
Webhook self-signed certificate issues for Virtual Cluster


crazywill:
In #145 and #161, we use self-signed certificate for ValidatingWebhookConfiguration. However, when vc-manager has multiple replicas,  every vc-manager will generate a new ValidatingWebhookConfiguration and delete old ValidatingWebhookConfiguration.  That's will case the webhook raise certificate error except for the latest vc-manager pod.
 
`https://virtualcluster-webhook-service.kube-system.svc:9443/validate-tenancy-x-k8s-io-v1alpha1-virtualcluster?timeout=30s": x509: certificate signed by unknown authority`

vincent-pli:
@crazywill 
Just curious, why `vc-manager` has multiple replicas? it's a `controller`, if there are more than one replicas, everyone will try to handle same object in its reconcile, that's not acceptable.

If you want to handle heavy load in the `controller`, you should increase the number of concurrent thread for the `reconcile` rather than make more replicas.

crazywill:
@vincent-pli Thank you for your reply.
As a controller, vc-manager runs in leader election mode, so it works well while having multiple replicas. But as a webhook, every replicas use its own caBundle, only the latest one can handle request.

vincent-pli:
@crazywill I'm afraid you are right, I try to fix it but I do not want to change too much.
Seems `controller-runtime` consider the case, see here: https://github.com/kubernetes-sigs/controller-runtime/issues/356

PR #194, opened 2021-07-22
✨ Add webhook framework

jichenjc:
**What this PR does / why we need it**:

enabled a set of manifests and webhook configuration, so the webhook sevice
will run on 9443 of capn-controller-manager , in case we edit the nestedcluster ,e.g 

```
kubectl patch --type=merge nestedcluster nestedcluster-sample -p '{"spec":{"controlPlaneEndpoint":{"port":6444}}}'
```

we will have following error
```
....
rolPlaneEndpoint:v1alpha4.APIEndpoint{Host:"localhost", Port:6444}}, Status:v1alpha4.NestedClusterStatus{Ready:false}}: Nestedclusters spec field is immutable. Please create a new resource instead. Ref doc: https://cluster-api.sigs.k8s.io/tasks/change-machine-template.html

```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
part of #184

PR #217, opened 2021-09-15
🐛 Fix webhook issue when there are multiple replication of controller

vincent-pli:
#215 

Fixes #

### DIFF
--- a/virtualcluster/cmd/manager/main.go
+++ b/virtualcluster/cmd/manager/main.go
@@ -140,11 +140,12 @@ func main() {
 	}
 
 	if enableWebhook == true {
-		log.Info("setting up webhooks")
-		if err := webhook.AddToManager(mgr, mgrOpt.CertDir); err != nil {
-			log.Error(err, "unable to register webhooks to the manager")
-			os.Exit(1)
+		webhookHandler := &webhook.WebhookHandler{
+			Manager: mgr,
+			CertDir: mgrOpt.CertDir,
 		}
+
+		mgr.Add(webhookHandler)
 	}
 
 	// Start the Cmd
--- a/virtualcluster/pkg/webhook/webhook.go
+++ b/virtualcluster/pkg/webhook/webhook.go
@@ -17,17 +17,33 @@ limitations under the License.
 package webhook
 
 import (
+	"context"
+
 	"sigs.k8s.io/controller-runtime/pkg/manager"
 )
 
 // AddToManagerFuncs is a list of functions to add all Controllers to the Manager
 var AddToManagerFuncs []func(manager.Manager, string) error
 
+type WebhookHandler struct {
+	Manager manager.Manager
+	CertDir string
+}
+
+func (wh *WebhookHandler) Start(context context.Context) error {
+	addToManager(wh.Manager, wh.CertDir)
+	return nil
+}
+
+func (wh *WebhookHandler) NeedLeaderElection() bool {
+	return true
+}
+
 // AddToManager adds all Controllers to the Manager
 // +kubebuilder:rbac:groups=admissionregistration.k8s.io,resources=mutatingwebhookconfigurations;validatingwebhookconfigurations,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=secrets,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups="",resources=services,verbs=get;list;watch;create;update;patch;delete
-func AddToManager(m manager.Manager, certDir string) error {
+func addToManager(m manager.Manager, certDir string) error {
 	for _, f := range AddToManagerFuncs {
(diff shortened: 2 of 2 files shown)


christopherhein (changes requested):
Rather than implementing this way can we maybe make this about removing the self-signing of certificates and setup the kustomize configs to auto annotate the controller to use a cert-manager based certificate like we get with normal Kubebuilder/controller runtime deployments?

vincent-pli:
@christopherhein  I'm OK for both solution.
but `cert-manger` itself is a big application and consist of many CRDs to handle its operation, is that a good idea to expect user install `cert-manager` just to handle admission webhook TLS certificate and CA?

christopherhein:
Yes, but the idea behind moving away from the vc manager auto publishing the validatingwebhookconfiguration it's self and delegating that to the configuration will make this controller work "properly" eg as you found you can't have multiple replicas cause it independently deploys separate self-signed certs, we need a layer above at provisioning that writes the cert for us so we can just use it. By moving to `cert-manager` as a default means that someone could inject their own certificate management in the same phase and not need to care about using cert-manager. So this opens up more options for the user.

kfox1111:
+1 to cert-manager. Better to rely on a project designed to handle certs for handling certs. And, cluster api already depends on it.

christopherhein:
We don't need to force users to use cert-manager but we should expect that if they choose not to use it that they manage certificates themselves.

cc @Fei-Guo @charleszheng44

christopherhein:
I would take inspiration from a default scaffolded kubebuilder project or even how the capn base project is setup with the configs/ directory it will use kustomize to auto-inject the right information but allows a user to also "opt out" of using cert-manager.

Fei-Guo:
I agree with building cert-manager integration for cert management in CAPN. Coming back to the fix, I think it is ok to have self-cert as long as it works without breaking fundamental architecture. We just don't recommend people to handle cert this way and provide cert-manager option next.

christopherhein:
I think the no-op that this is causing users is that if they use the `ENABLE_WEBHOOK` flag it will auto deploy the ValidatingWebhookConfiguration from each replica of the vc-manager that they run, meaning the certificate that is supplied to the webhook might only work for one of the running replicas in the cluster and your webhook request might hit the wrong instance causing it to return `https://virtualcluster-webhook-service.kube-system.svc:9443/validate-tenancy-x-k8s-io-v1alpha1-virtualcluster?timeout=30s": x509: certificate signed by unknown authority`. 

So right now the architecture doesn't allow vc-manager to run as more than one replica for higher availability.

vincent-pli:
@christopherhein thanks for your comments.
Actually with this fix, the issue you presented will gone.
In HA case, only the replication who win the `leader election` will create `admission webhook` and sign the cert.
When fail-over occurred, the new winner will delete the older `webhook` and re-sign the cert, so the original issue will gone, we can have HA.

For `cert-manager`, I agree we can make integration at CAPN level, should be next step.

christopherhein:
Hmm, are you sure? I am 90% sure when you use the `LeaderElectedRunnable` interface like you are in this PR you are only exposing the webhook handlers on the leader meaning, running multiple replicas doesn't help you scale the webhook horizontally like in a traditional webhook deployment. So this doesn't actually make it HA you now just can support hot standby controllers/webhook handler which isn't the same.

christopherhein:
Maybe you could provide some samples showing a curl to the webhook endpoint on a non-leader running in a cluster then show the same against the leader and that would help me.

crazywill:
I test in my env, it doesn't work because virtualcluster-webhook-service use selector: **virtualcluster-webhook=true**.  
How about setting **Endpoints** instead of using selector. I will try and test it.

christopherhein:
You should be able to port-forward and then curl the pod on the webhook port and see if it's listening.

crazywill:
It works when i delete the **label selector** in virtualcluster-webhook-service and make winner create **Endpoints** itself (delete first if exists). 

vc-manager has three replication, and endpoint points to the winner pod as expected.
image

image

vincent-pli:
Thanks @christopherhein, I 100% use the `LeaderElectedRunnable` : )

The fix just resolve the conflict of cert generated but I forget the `SVC`, it act a role of load balance so will redirect validation request to a replication who has not set the http server(loser of leader election).

There is several solution to resolve the problem(as @crazywill mentioned) but I think the easiest way is to adopt `cert-manager`.

christopherhein:
Agreed, how do you want to handle this, did you want to use this PR to move over to cert-manager based?

christopherhein:
Sounds good, thank you!

Make sure to take a look at the CAPN config directory for inspiration and direction so that it will auto-deploy everything for you.

christopherhein:
Check out https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/194 for when that was implemented.

PR #219, opened 2021-09-30
⚠️ syncer: clean opaqued keys before construct super cluster object

zhuangqh:

prevent tenant from using super cluster annotation/labels

Fixes #

### DIFF
--- a/virtualcluster/pkg/syncer/conversion/helper.go
+++ b/virtualcluster/pkg/syncer/conversion/helper.go
@@ -25,22 +25,25 @@ import (
 	"fmt"
 	"strings"
 
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-
 	"github.com/pkg/errors"
 	v1 "k8s.io/api/core/v1"
 	v1scheduling "k8s.io/api/scheduling/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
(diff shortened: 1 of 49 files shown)


virtualcluster/pkg/syncer/conversion/helper.go
```diff
@@ -163,7 +192,28 @@ func BuildMetadata(cluster, vcns, vcname, targetNamespace string, obj client.Obj
...
-func BuildSuperMasterNamespace(cluster, vcName, vcNamespace, vcUID string, obj client.Object) (client.Object, error) {
+func (c *objectConversion) CleanOpaqueKeys(vc *v1alpha1.VirtualCluster, keyMap map[string]string) {
+	var exceptionsList []string
+	if vc != nil {
+		exceptions := sets.NewString()
+		exceptions.Insert(vc.Spec.OpaqueMetaPrefixes...)
+		exceptions.Insert(constants.DefaultOpaqueMetaPrefix, constants.DefaultTransparentMetaPrefix)
+		exceptionsList = exceptions.UnsortedList()
+	}
+
```
Fei-Guo, line 192: Should we check the length of the list and return directly if it is 0 (although in practice it is not 0)?

zhuangqh: we should also check the opaquedomain here, cannot directly return even if exceptionlist is 0

Fei-Guo:
Most of the code changes are refactoring changes. The added logic makes sense for preventing tenant from abusing certain keys maintained by the super cluster.

/lgtm
/approve

PR #223, opened 2021-10-13
use namespace as kubeletclient cert cn because vn-agent expects that for kubectl exec

jzhoucliqr:

for kubectl exec/log/port forward, vn-agent needs to know the namespace in supercluster, currently it assumes the kubelet-client cert that virtual apiserver uses have the CN as the namespace, so add this hack to make it work. 

Also removed the flag 'kubelet-certificate-authority' in apiserver args because with that virtual apiserver is trying to validate the cert that vn-agent is using, which will fail because the cert doesn't have the pod IP as SAN. 

Fixes #

### DIFF
--- a/controlplane/nested/certificate/helpers.go
+++ b/controlplane/nested/certificate/helpers.go
@@ -75,10 +75,12 @@ func NewAPIServerCrtAndKey(ca *KeyPair, clusterName, clusterDomainArg, apiserver
 
 // NewAPIServerKubeletClientCertAndKey creates certificate for the apiservers to connect to the
 // kubelets securely, signed by the ca.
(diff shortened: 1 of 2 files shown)


christopherhein:
What do we want/need to do with this PR @jzhoucliqr and @charleszheng44 ?

christopherhein:
@jzhoucliqr would you be able to still rebase this, sorry for the super long delay but this PR makes sense to me.

jzhoucliqr:
> @jzhoucliqr would you be able to still rebase this, sorry for the super long delay but this PR makes sense to me.

Rebased. Yeah It's a long time. Almost forgot about this.

christopherhein:
thank you and sorry! @jzhoucliqr 

/lgtm
/approve

PR #226, opened 2021-10-15
:book: Add Q&A to explain the difference between Loft vCluster and VirtualCluster

Fei-Guo:
VirtualCluster

Since the question is being frequently asked, let me add the answer to the Q&A section.

### DIFF
--- a/virtualcluster/README.md
+++ b/virtualcluster/README.md
@@ -111,6 +111,24 @@ maintains the one to one mapping between a virtual node in tenant control plane 
 in the super cluster. It preserves the Kubernetes API compatibility as closely as possible. Additionally, 
 it provides fair queuing to mitigate tenant contention.
 
+### Q: What is the difference between VirtualCluster and Loft vCluster?
+
+Although VirtualCluster and vCluster share similar architecture such that each tenant is given
+a dedicated control plane to achieve strong isolation, their target use cases are different.
+The difference leads to divergent implementations which impact API compatibility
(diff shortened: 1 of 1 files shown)


LukasGentele:
@Fei-Guo As mentioned in the comments above, there definitely are a few differences but the n:n vs n:1 namespace translation is the core architectural difference and I think that should be clearly articulated but I think both approaches are valid architectural choices whereas this document suggests that n:1 namespace translation has API "compatibility issues" which is not correct. n:1 namespace translation can cover everything that n:n namespace translation does, it just requires additional rewriting within the syncer for things that have a namespace selector, for example.

Another notable difference is k8s vs k3s as API server / control plane. While we plan to add k8s itself in vcluster, currently it's only k3s.

I think some of the differences may disappear over time when projects expand or they may get larger depending on where things are heading and I agree with you that that's not really an issue. I assume there will be a lot more virtual cluster implementations in the future just like there are multiple policy engines, storage backends and network plugins for kubernetes. I mean that's a sign of a healthy ecosystem and probably the biggest achievement of k8s and the CNCF to allow for all of these parallel solutions to exist and thrive and to allow that everyone can pick and choose what works best for them and for their use cases.

Fei-Guo:
Fair enough. I will close the PR since we should not judge which implementation is better given that two implementations provide the same user experience.  Good to hear your progress on resolving the compatibility issues. 

As you may know, VirtualCluster is shifting the focus to multiple super cluster support/management and we have no plan to abstract the syncer layer and I don't see the need neither.

PR #232, opened 2021-11-01
🐛 VC-Syncer - remove check for optional secret Volumes on Pods

### LINKED ISSUE #231: VC-Syncer fails to sync pods with optional secrets

Apply the following to a virtual cluster:

```
apiVersion: v1
kind: Pod
metadata:
  name: optional-secret-test
spec:
  containers:
  - image: busybox
    name: test
    command: ['sh', '-c', 'while true; do echo "alive" && sleep 3600; done']
  restartPolicy: Always
  volumes:
    - name: optional-password
      secret:
        optional: true
        secretName: optional-password
```

Expected it to schedule but the pod is stuck in pending state.

Events:
```
  Type     Reason        Age                From       Message
  ----     ------        ----               ----       -------
  Warning  FailedCreate  7s (x12 over 17s)  vc-syncer  Error creating: failed to get service account secret from cluster test-cluster cache: failed to get vSecret default/optional-password: Secret "optional-password" not found
```

[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- kind v0.11.1
- super cluster - Version: GitVersion:"v1.19.13"
- tenant - Version: GitVersion:"v1.18.20"

/kind bug

lukeweber:

Remove a check for optional secrets in pod syncer. The pod resource syncer checks that all secrets defined in volumes in the pod spec exist in the virtual cluster. The intent is to ensure that service account tokens exist before the pod is scheduled. The minimal change is to only verify secrets that are marked as required. If it's optional, we remove it from the check.

We should confirm that ignoring optional doesn't compromise any of the security assumptions of the current model in terms of us making sure a service account token is properly synced before a pod is created.

We do already enforce this behavior with syncer optionally setting AutomountServiceAccountToken on the sync'd pods spec:
-  [DisableServiceAccountToken](https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/1eb78b9401490286579ac1aab0a51656e4547ae6/virtualcluster/pkg/syncer/apis/config/types.go#L51) 
- [syncer mutate using DisableServiceAccountToken](https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/2aa1c8cf8533419e790a56c2b68ba2fdfe04f9ae/virtualcluster/pkg/syncer/conversion/mutate.go#L391)

Note: The negative case is already covered with tests that verify that a ServiceAccount secret not existing will cause the syncer to fail. I'm open to adding to the tests if we feel it's useful, but my judgement was to not add them.

Fixes #231

### DIFF
--- a/virtualcluster/pkg/syncer/resources/pod/dws.go
+++ b/virtualcluster/pkg/syncer/resources/pod/dws.go
@@ -31,6 +31,7 @@ import (
 	"k8s.io/apimachinery/pkg/util/sets"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
+	"k8s.io/utils/pointer"
 
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/constants"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/conversion"
@@ -222,7 +223,7 @@ func (c *controller) reconcilePodCreate(clusterName, targetNamespace, requestUID
 func (c *controller) findPodServiceAccountSecret(clusterName string, pPod, vPod *v1.Pod) (map[string]string, error) {
 	mountSecretSet := sets.NewString()
 	for _, volume := range vPod.Spec.Volumes {
-		if volume.Secret != nil {
+		if volume.Secret != nil && !pointer.BoolDeref(volume.Secret.Optional, false) {
 			mountSecretSet.Insert(volume.Secret.SecretName)
 		}
 	}
--- a/virtualcluster/pkg/syncer/resources/pod/dws_test.go
+++ b/virtualcluster/pkg/syncer/resources/pod/dws_test.go
@@ -81,6 +81,15 @@ func tenantPod(name, namespace, uid string) *v1.Pod {
 						},
 					},
 				},
+				{
+					Name: "i-do-not-exist-optional",
+					VolumeSource: v1.VolumeSource{
+						Secret: &v1.SecretVolumeSource{
+							SecretName: "i-do-not-exist",
+							Optional:   pointer.Bool(true),
+						},
+					},
+				},
 			},
 		},
 	}
@@ -170,6 +179,15 @@ func superPod(clusterKey, vcName, vcNamespace, name, namespace, uid string) *v1.
 					VolumeSource: v1.VolumeSource{
 						Secret: &v1.SecretVolumeSource{
 							SecretName: testSuperServiceAccountTokenSecretName,
+						},
+					},
+				},
+				{
+					Name: "i-do-not-exist-optional",
+					VolumeSource: v1.VolumeSource{
+						Secret: &v1.SecretVolumeSource{
+							SecretName: "i-do-not-exist",
+							Optional:   pointer.Bool(true),
 						},
 					},
 				},


lukeweber:
Apologies for excessive pushes. Was trying to get my verified commit to work correctly.

christopherhein:
@Fei-Guo or @charleszheng44 have either of you run into this before? We ran into it in some testing and want to make sure this doesn't break anything you have been using.

Fei-Guo:
I don't know the use case of this `optional` field. Do you mind telling us when this field is used? Changing this field requires Pod recreation, what is the difference between flipping a boolean compared to adding a secret in volume?

lukeweber:
Per the original issue #231 - If someone creates a pod with an optional secret, and the secret is never created, the following should be a valid / runnable pod. However, with the current check that assume all secrets are service accounts and thus required to be synced before scheduling the pod in the super cluster, the pod will never be synced to the super cluster by vc-syncer which is incorrect.

Example pod
```yaml
apiVersion: v1
kind: Pod
metadata:
  name: optional-secret-test
spec:
  containers:
  - image: busybox
    name: test
    command: ['sh', '-c', 'while true; do echo "alive" && sleep 3600; done']
  restartPolicy: Always
  volumes:
    - name: optional-password
      secret:
        optional: true
        secretName: optional-password
```

Fei-Guo:
Yes, I understand the logic but don't quite understand the use case and I am curious if someone is really using this field. Is this from a conformance test?

christopherhein:
> Yes, I understand the logic but don't quite understand the use case and I am curious if someone is really using this field. Is this from a conformance test?

It's a bit different, @lukeweber provided that pod spec which will fail to be synced to the super cluster because of the optional field, our current checks will loop through all secrets and will fail if one doesn't exist which breaks the contract for what the `optional` field is supposed to be doing, making it so any pod with an optional secret won't ever be synced to the super cluster.

Fei-Guo:
If the flag is true and secret does not exists, we should be ok. The problem is when the flag is true and the secret exists, your change allows a race such that the user created secret and Pod in order in VC but the syncer creates a Pod in super without the secret volume because the syncer cannot enforce mutual synchronization order for any objects in VC. In my option, I'd rather claim we don't support this field. What do you think?

christopherhein:
> If the flag is true and secret does not exists, we should be ok. The problem is when the flag is true and the secret exists, your change allows a race such that the user created secret and Pod in order in VC but the syncer creates a Pod in super without the secret volume because the syncer cannot enforce mutual synchronization order for any objects in VC. In my option, I'd rather claim we don't support this field. What do you think?

I don't think that is right in how it was implemented upstream prior to this PR, cause right now that code only checks for if the ServiceAccount Token Secret exists in the super before syncing, all other secret volumes are ignored since this can be handled as eventually consistent expect a ServiceAccount token Secret and a ServiceAccount token Secret shouldn't ever be `Optional`. See the linked code below for where we skip over non-SA token secrets.

https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/0e68290509fc7b9fe9cfbc415458531ede1a309c/virtualcluster/pkg/syncer/resources/pod/dws.go#L241-L243

I personally think this is a feature we should support given it's scoped pretty specific to optional non-SA secrets. Then again I might be misunderstanding the code.

Fei-Guo:
I agree we have the tiny race problem in general for non SA secrets or other volume sources with optional fields being true.

PR #241, opened 2022-02-02
🐛 Fix non-leader crashloop

m-messiah:
Syncer non-leader pod could be stuck in CrashLoopBackOff because of /healthz unavailable when LeaderElection does not hit the leader callback.

The PR moves /healthz server, pprof server and /metrics listeners outside of leader election logic, allowing them to be available always.

### DIFF
--- a/virtualcluster/cmd/syncer/app/server.go
+++ b/virtualcluster/cmd/syncer/app/server.go
@@ -122,7 +122,7 @@ func Run(cc *syncerconfig.CompletedConfig, stopCh <-chan struct{}) error {
 	defer cancel()
 
 	// Prepare a reusable runCommand function.
-	run := startSyncer(ctx, ss, cc, stopCh)
+	run := startSyncer(ss, stopCh)
 
 	go func() {
 		select {
@@ -130,6 +130,22 @@ func Run(cc *syncerconfig.CompletedConfig, stopCh <-chan struct{}) error {
 			cancel()
 		case <-ctx.Done():
 		}
+	}()
+
+	go func() {
+		// start a pprof http server
+		klog.Fatal(http.ListenAndServe(":6060", nil))
+	}()
+
+	go func() {
+		// start a health http server.
+		mux := http.NewServeMux()
+		healthz.InstallHandler(mux)
+		klog.Fatal(http.ListenAndServe(":8080", mux))
+	}()
+
+	go func() {
+		ss.ListenAndServe(net.JoinHostPort(cc.Address, cc.Port), cc.CertFile, cc.KeyFile)
 	}()
 
 	if cc.LeaderElection != nil {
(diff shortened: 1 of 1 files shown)


virtualcluster/cmd/syncer/app/server.go
```diff
@@ -132,6 +132,25 @@ func Run(cc *syncerconfig.CompletedConfig, stopCh <-chan struct{}) error {
...
+		klog.Fatal(http.ListenAndServe(":6060", nil))
+	}()
+
+	var leaderElector *leaderelection.LeaderElector
+	go func() {
+		// start a health http server.
+		mux := http.NewServeMux()
+		healthz.InstallHandler(mux)
+		healthz.InstallReadyzHandler(mux, healthz.NamedCheck("leader", func(r *http.Request) error {
+			if leaderElector != nil && !leaderElector.IsLeader() {
```
munnerz, line 146: This will probably cause issues during a rolling upgrade. Pods must transition to ‘Ready’ to allow the upgrade to proceed when managed by the Deployment controller.

Readiness shouldn't be used to indicate whether instances are leaders 😄

m-messiah: Ok, I removed it for this PR. To be followed up later

virtualcluster/cmd/syncer/app/server.go
```diff
@@ -132,6 +132,22 @@ func Run(cc *syncerconfig.CompletedConfig, stopCh <-chan struct{}) error {
 		}
 	}()
 
+	go func() {
+		// start a pprof http server
+		klog.Fatal(http.ListenAndServe(":6060", nil))
```
munnerz, line 137: I don't think this is an issue caused in this PR, but what area of code actually registers pprof endpoints with this HTTP server/muxer? 🤔

virtualcluster/cmd/syncer/app/server.go
```diff
@@ -132,6 +132,22 @@ func Run(cc *syncerconfig.CompletedConfig, stopCh <-chan struct{}) error {
...
+
+	go func() {
+		// start a health http server.
+		mux := http.NewServeMux()
+		healthz.InstallHandler(mux)
+		klog.Fatal(http.ListenAndServe(":8080", mux))
+	}()
+
+	go func() {
+		ss.ListenAndServe(net.JoinHostPort(cc.Address, cc.Port), cc.CertFile, cc.KeyFile)
```
munnerz, line 148: Aha, digging into the syncer code, it seems like this is actually for *metrics* serving rather than for webhooks: https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/e5ca44baf3040c0f5cbeed7b20f2fd3c2c0b4d72/virtualcluster/pkg/syncer/syncer.go#L261

Given that the syncer isn't running when not elected (`s.Run` has not been called), I think this will just be an empty page of metrics which is what we want & shouldn't require any additional work in prometheus rules...

virtualcluster/cmd/syncer/app/server.go
m-messiah: This is the pprof behaviour - it handles the default muxer: https://pkg.go.dev/net/http/pprof

munnerz: Nice! Thanks 😄

munnerz:
/lgtm

You'll need an approver from https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/OWNERS or https://github.com/orgs/kubernetes-sigs/teams/cluster-api-provider-nested-maintainers to add the `/approve`'d label!

virtualcluster/cmd/syncer/app/server.go
m-messiah: As the whole work will be done in the leader, having metrics from the non-leader instance is not a big deal, as they are eventually reset

christopherhein (approved):
/lgtm
/approve

PR #243, opened 2022-02-03
🌱 Configurable wait time on VirtualCluster Controller for control plane bring up

m-messiah:
Today the manager will timeout after 180seconds and stop trying to deploy the control plane. This causes problems if someone is trying to create a cluster when there is super cluster issues happening.
The  PR allows to configure timeout with `--provisioner-timeout` flag for manager.

### DIFF
--- a/virtualcluster/cmd/manager/main.go
+++ b/virtualcluster/cmd/manager/main.go
@@ -20,6 +20,7 @@ import (
 	"flag"
 	"fmt"
 	"os"
+	"time"
 
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/apis"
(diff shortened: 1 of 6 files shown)


virtualcluster/cmd/manager/main.go
```diff
@@ -64,6 +65,7 @@ func main() {
 	flag.BoolVar(&versionOpt, "version", false, "Print the version information")
 	flag.BoolVar(&disableStacktrace, "disable-stacktrace", false, "If set, the automatic stacktrace is disabled")
 	flag.BoolVar(&enableWebhook, "enable-webhook", false, "If set, the virtualcluster webhook is enabled")
+	flag.Int64Var(&provisionerTimeout, "provisioner-timeout", 180, "The timeout for provision control-plane statefulsets")
```
atosatto, line 68: I would consider using `flag.DurationVar` and setting the default to `600 * time.Second` as the current configuration.

m-messiah: the current for Native is 180, the current for aliyun is 600.

I thought about duration, but we use kubeutil.WaitStatefulSetReady, which accepts int64 and there is no reason to do a double conversion (except for changing kubeutil.WaitStatefulSetReady signature, which is used in several places)

munnerz: I am +1 with using Duration personally - whilst the underlying machinery may use an int or whatever else, the flags can act as a translation layer/user interface. By using a duration, users can use more natural language like `5m` or `5m30s` etc to indicate the timeout. Just my 2c :)

m-messiah: Updated to Duration

munnerz (approved):
/lgtm

christopherhein (approved):
/approve
/lgtm

PR #245, opened 2022-02-08
🐛 Fix a corner case that cluster gets partially deleted and ends up w…

weiling61:
…ith a nil cluster pointer

In some rare cases, cluster map in syncer got corrupted and returns nil cluster, and syncer cannot recover from it and keep crashing. Do a nil pointer check

Fixes #


christopherhein (approved):
/lgtm
/approve

PR #248, opened 2022-02-10
🐛 When using ClusterIP based services we try to double create the service causing error messages

m-messiah:
This PR skips the creation of clusterIP apiserver service when it must be already created in the CreateVirtualCluster method.

There is also a formatting change which probably accidentally was introduced earlier


christopherhein (approved):
/lgtm
/approve

PR #252, opened 2022-02-15
✨Enabling Metrics for vn-agent

### LINKED ISSUE #251: ✨ Metrics for vn-agent
## User Story

As an operator I would like to export metrics for vn-agent.

## Detailed Description

vn-agent does not export metrics at this point. A metrics server is to be enabled by default on the address `:9100` and allowed to be configured by the cli option `--metrics-addr` options. The metrics server can be disabled by cli option `--enable-metrics` to `false`. The metrics are default enabled.

`vn-agent` should expose the below metrics

- Instrumented API calls to super api / Kubelet. 
- Count requests with respect to host, tenantName, tenantNamespace, action and responseCode.

/kind feature

srirammageswaran8:

The pr enabled metrics server for vn-agent instrumenting all the API Calls to super API Server or Kubelet and counting all the API Calls with respect to host, tenantName, tenantNamespace, action and responseCode.

**Which issue(s) this PR fixes**
Fixes #251

### DIFF
--- a/virtualcluster/pkg/vn-agent/server/metrics.go
+++ b/virtualcluster/pkg/vn-agent/server/metrics.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


virtualcluster/pkg/vn-agent/server/metrics.go
```diff
@@ -0,0 +1,124 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
christopherhein, line 2: Can you update the copyright on this to 2022.

```suggestion
Copyright 2019 The Kubernetes Authors.
```

christopherhein (changes requested):
This is looking really good, just one comment about a copyright that should be updated.

cc @Fei-Guo @charleszheng44 would either of you want to take a look at this before merging?

PR #254, opened 2022-02-25
✨ Adding Support for Mutating DNSPolicy

christopherhein:
This adds a feature gate that allows us to mutate the pods dnsPolicy using a label, removing the mutation that disallows setting `ClusterFirst` DNS.


virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -172,7 +173,10 @@ func PodMutateDefault(vPod *v1.Pod, saSecretMap map[string]string, services []*v
 		if err != nil {
 			return err
 		}
-		mutateDNSConfig(p, vPod, vc.Spec.ClusterDomain, nameServer)
+		allowDNSPolicyMutation := featuregate.DefaultFeatureGate.Enabled(featuregate.TenantAllowDNSPolicyMutation)
+		if _, set := vPod.GetLabels()[constants.TenantDisableDNSPolicyMutation]; !allowDNSPolicyMutation && !set {
```
Fei-Guo, line 177: The naming looks confusion. We call `mutateDNSConfig` when `allowDNSPolicyMutation` is false?

christopherhein: good point, let me change that.

virtualcluster/pkg/syncer/conversion/mutate.go
```diff
@@ -172,7 +173,10 @@ func PodMutateDefault(vPod *v1.Pod, saSecretMap map[string]string, services []*v
 		if err != nil {
 			return err
 		}
-		mutateDNSConfig(p, vPod, vc.Spec.ClusterDomain, nameServer)
+		allowTenantDNSPolicy := featuregate.DefaultFeatureGate.Enabled(featuregate.TenantAllowDNSPolicy)
+		if _, set := vPod.GetLabels()[constants.TenantDisableDNSPolicyMutation]; !allowTenantDNSPolicy && !set {
```
Fei-Guo, line 177: I think it should be `!(allowTenantDNSPolicy && set)` ?

christopherhein: Isn't this logic the same? `&&` covers that, unless you mean stylistically.

Fei-Guo: The mutate call is skipped if someone sets the constants.TenantDisableDNSPolicyMutation label even if the feature gate is disabled.

christopherhein: Oh duh 🤦

weiling61 (approved):
/lgtm

Issue #240, opened 2022-01-17
Performance and Resource Contention in Multiple Tenant Environment


weiling61:
**User Story**
Would like to add:
- Plugin to validate and limit total resource usage of a virtual cluster (tenant)
- Add per virtual cluster working queue and workers in mccontroller to prevent contention among tenants.

**Detailed Description**
Performance and resource contention among tenants are not fully addressed by virtual cluster. More monitoring or enforcement logics are needed to control total resource usage of a tenant.

### **Resource Quota:**
In addition to per namespace based admission quota control, overall resource quota needs to be checked for each tenant (virtual cluster) across multiple tenant’s namespaces.  This is to limit tenant’s overall resource usage in a multi-tenant environment. 
 
We would like to propose adding resource creation validation logics in Pod syncer.  Pod Syncer is a perfect location to perform the monitoring, validation, and enforcement. For example before a pod is going to be created in super cluster in “reconcilePodCrete()”, a validation logic can be injected. If the validation fails, the pod object is kept in tenant virtual cluster with pending state. The pod will be later reconciled and got synced to super cluster once resources become available.
 
Given quota enforcement is proprietary to each cloud environment, customers may choose their own way to perform the validation, we would like to add a plugin in Pod Syncer to allow customers to use their own validation logics. Customers can register the validation plugin at Syncer startup. Pod syncer then calls the plugin in “reconcilePodCrete()” before creating pod in super cluster.
 
In addition, a new label can be added in pod that does not pass the validation. Pod Syncer can bypass re-reconciling of the pods to reduce un-necessary work load. A separate user logic can be added to take the label off from pending pods when resource become available.

### **Syncer Performance Contention:**
Currently, mccontroller uses MaxConcurrentReconciles =3 of workers to run syncing for all tenants’ virtual clusters. It would be nice to have per tenant (virtual cluster) work queue and per tenant workers. This is to prevent noisy tenant from interfere other tenant’s syncing job. This also provides the ability to serialize pod syncing for each tenant, which is critical for certain use cases --- for example to get accurate resource usage status for resource quota validation.

christopherhein:
Hey @weiling61 thanks for filing this. I think there might be two parts to this issue that we can probably untease. 1. a plugin architecture that would allow someone to customize the builtin resource syncers For now I'm going to focus on the plugin architecture. (I believe we have other outlets to improve performance of the max workers with how the fair queue is setup per tenant)

[...]

PR #255, opened 2022-03-08
🌱Add validation plugin and add tenant quota validation before creating a pod

weiling61:
- Add validation plugin framework and interface
- Add tenant quota validation utilities
- Add validation plugin handling in pod syncer 

Following is the proposal for this feature:
https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/docs/proposals/20220206-quota.md

Following issue was originally raised:
Performance and Resource Contention in Multiple Tenant Environment #240

Fixes # Enhancement


weiling61:
@Fei-Guo @christopherhein

Fei-Guo:
Can we just have ValidationPluginInterface defined with three methods
```
type ValidationPluginInterface interface {
	Validation(client.Object, string) bool
	GetTenantLocker(string) *Tenant
	Enabled() bool
}
```
so the changes to the Pod dws is straightforward.  You can put rest of the code in 
resources/validationplugin/validationplugin.go to a separate file called quotavalaidationPlugin which implements the interface. I would suggest to name the plugin as `QuotaValidationPlugin` explicitly.

Also, please consider to move the validationplugin directory under `resources/pod/` since the plugin works for Pod resource only.

For the plugin definition, can you just reuse the one that has been added in the util package?
sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/util/plugin/plugin.go

Fei-Guo:
Can we just have ValidationPluginInterface defined with three methods
```
type ValidationPluginInterface interface {
	Validation(client.Object, string) bool
	GetTenantLocker(string) *Tenant
	Enabled() bool
}
```
so the changes to the Pod dws is straightforward.  You can put rest of the code in 
resources/validationplugin/validationplugin.go to a separate file called quotavalaidationPlugin which implements the interface. I would suggest to name the plugin as `QuotaValidationPlugin` explicitly.

Also, please consider to move the validationplugin directory under `resources/pod/` since the plugin works for Pod resource only.

For the plugin definition, can you just reuse the one that has been added in the util package?
sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/util/plugin/plugin.go

weiling61:
@Fei-Guo Thanks for the review. Updated according to your review comments

weiling61:
@Fei-Guo updated according to your review comments.

Fei-Guo:
What happened to the virtualcluster/README.MD file?

PR #256, opened 2022-03-09
✨ Supporting Super Cluster Timeout Overrides

christopherhein:
The default hardcoded 30 second timeout for the super cluster isn't enough under some scenarios, this allows that to be additionally supplied from a CLI flag but still uses the standard default if not supplied.

### DIFF
--- a/virtualcluster/pkg/syncer/apis/config/types.go
+++ b/virtualcluster/pkg/syncer/apis/config/types.go
@@ -70,6 +70,9 @@ type SyncerConfiguration struct {
 
 	// Super cluster rest config
 	RestConfig *rest.Config
+
+	// The maximum length of time to wait before giving up on a server request. A value of "" means use default.
+	Timeout string
 }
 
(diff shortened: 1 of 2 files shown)


virtualcluster/pkg/syncer/apis/config/types.go
```diff
@@ -70,6 +70,9 @@ type SyncerConfiguration struct {
 
 	// Super cluster rest config
 	RestConfig *rest.Config
+
+	// The maximum length of time to wait before giving up on a server request. A value of "" means use default.
```
Fei-Guo, line 74: Can you explain the unit of timeout in the comment (second or millisecond, etc.)?

Fei-Guo (commented):
LG

One nit.

virtualcluster/pkg/syncer/apis/config/types.go
christopherhein: Yes, I can follow up with something, the user can supply anything from `5s`, `5m`, `1h`, anything that `time.ParseDuration` accepts.

PR #260, opened 2022-03-16
🌱 Increase log level threshold for "giving up" messages

m-messiah:
It is useless to have these messages to spam in logs in case the syncer is doing nothing.
Let's move them to the next verbosity level as other debug logs.


christopherhein:
I believe there is an outage causing prow to struggle. https://kubernetes.slack.com/archives/C7J9RP96G/p1647442505258529

PR #266, opened 2022-03-28
transfer k8s api to meet new version

vincent-pli:
The latest k8s version adopted by `Kind` is `v1.23`, and there are several API is deprecated and not support in the latest version, include:
- MutatingWebhookConfiguration and ValidatingWebhookConfiguration from `v1beta1` to `v1`
- apiextensions.k8s.io from `v1beta1` to `v1`

So update some code snippet to make it works with the new k8s version


christopherhein:
ok-to-test

christopherhein (approved):
/lgtm
/approve

PR #267, opened 2022-03-28
🌱 Switching to golangci-lint via actions

christopherhein:
It seems golangci-lint in prow has started to act up. I'm hoping moving to the standard way to run this as other projects like CAPI did https://github.com/kubernetes-sigs/cluster-api/pull/4374


weiling61 (approved):
/lgtm

PR #268, opened 2022-03-31
Remove deprecated annotation for SVC, it block the installation of etcd

vincent-pli:
With the k8s version 1.22+, the `etcd` cannot install success, since the original annotation is deprecated:
`service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"`
That's make the `coredns` cannot resove the pod(etcd pod) which is not ready.

See the issue:
https://github.com/coredns/coredns/issues/5159#issuecomment-1033726643

### DIFF
--- a/virtualcluster/config/sampleswithspec/clusterversion_v1_nodeport.yaml
+++ b/virtualcluster/config/sampleswithspec/clusterversion_v1_nodeport.yaml
@@ -103,9 +103,8 @@ spec:
       kind: Service
       metadata:
         name: etcd
-        annotations:
-          service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
       spec:
+        publishNotReadyAddresses: true
         type: ClusterIP
         clusterIP: None
         selector:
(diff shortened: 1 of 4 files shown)


vincent-pli:
@christopherhein

virtualcluster/config/sampleswithspec/clusterversion_v1_nodeport.yaml
```diff
@@ -103,9 +103,8 @@ spec:
       kind: Service
       metadata:
         name: etcd
-        annotations:
-          service.alpha.kubernetes.io/tolerate-unready-endpoints: "true"
       spec:
+        publishNotReadyAddresses: true
```
christopherhein, line 107: Do we need/want this for the `clusterversion_v1_loadbalancer.yaml` as well?

christopherhein:
The builds recently started failing cause of some change in test-infra, so this is likely going to be stuck for a bit while we work out that. I started this work here -https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/267 but stalled out with other work.

virtualcluster/config/sampleswithspec/clusterversion_v1_nodeport.yaml
vincent-pli: The file has been changed,  please check the change list. @christopherhein

christopherhein: 🤦 sorry about that!

christopherhein (approved):
/lgtm
/approve

PR #269, opened 2022-04-25
✨ Adding VN Agent PodIP Provider

christopherhein:

This adds the ability to use the `vn-agent` as a `DaemonSet` but uses vn-agent pod IPs as the routable address to configure with the vnode UWS'ed object. Because this interface is used for all syncing any updates to the node status object trigger a reconciliation. We handle vn-agent pod failures with status updates which triggers a resync.

> ⚠️  For anyone implementing this you should have a node level check for example `node-exporter` or an `initContainer` on the vn-agent which updates a condition on the Node or other part of the node object. By doing this the syncer will update all your VirtualClusters VNodes anytime the `vn-agent` is rescheduled, this can be done without rebinding.

No issue filed. 

### DIFF
--- a/virtualcluster/pkg/syncer/vnode/pod/provider.go
+++ b/virtualcluster/pkg/syncer/vnode/pod/provider.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package pod allows you to configure the syncer with vnodes backed by a vn-agent running without hostNetworking
+//
+// WARNING: For anyone implementing this you should have a node level check for example node-exporter or an
+// initContainer on the vn-agent which updates a condition on the Node or other part of the node object. By doing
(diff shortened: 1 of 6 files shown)


virtualcluster/pkg/syncer/vnode/pod/provider.go
```diff
@@ -0,0 +1,82 @@
...
+func (p *provider) GetNodeAddress(node *v1.Node) ([]v1.NodeAddress, error) {
+	var addresses []v1.NodeAddress
+	namespaceName := strings.Split(p.vnAgentNamespaceName, "/")
+	// TODO(christopherhein) Use NodeName informer index to make this more efficient.
+	pods, err := p.client.CoreV1().Pods(namespaceName[0]).List(context.TODO(), metav1.ListOptions{LabelSelector: p.vnAgentLabelSelector})
+	if err != nil || len(pods.Items) == 0 {
+		return addresses, fmt.Errorf("vn-agent pods could not be found %s", err)
+	}
+
+	pod := v1.Pod{}
```
Fei-Guo, line 66: You may only need a pointer here to avoid copying an object.

Fei-Guo (commented):
"⚠️ For anyone implementing this you should have a node level check for example node-exporter or an initContainer on the vn-agent which updates a condition on the Node or other part of the node object. By doing this the syncer will update all your VirtualClusters VNodes anytime the vn-agent is rescheduled, this can be done without rebinding."

Can you add the above in the code comment somewhere?

christopherhein:
Blocked on #270

virtualcluster/pkg/syncer/vnode/pod/provider.go
christopherhein: Good idea! and I will add the comments to the code.

christopherhein: Done! ✅

Fei-Guo:
One last nit, please change the copyright year for all touching files.

christopherhein:
@Fei-Guo ✅ done, just waiting for tests to pass again.

PR #270, opened 2022-04-25
🐛 Github Actions goreleaser fixes

christopherhein:

Since #267 was merged we hadn't tested golangci-lint w/ GitHub actions, this is attempting to fix the issues with the setup.

### DIFF
--- a/main.go
+++ b/main.go
@@ -27,8 +27,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientgoscheme "k8s.io/client-go/kubernetes/scheme"
-        "k8s.io/component-base/logs"
-        klog "k8s.io/klog/v2"
 	"k8s.io/klog/v2/klogr"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha4"
 	"sigs.k8s.io/cluster-api/feature"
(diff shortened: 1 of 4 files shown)


christopherhein:
cc @weiling61

christopherhein:
cc @Fei-Guo

main.go
```diff
@@ -60,6 +60,9 @@ var (
 )
 
 func init() {
+	logs.InitLogs()
+	defer logs.FlushLogs()
```
Fei-Guo, line 64: Is calling `defer` in the init function an intended change? This seems to changing the old behavior.

weiling61 (approved):
/lgtm

main.go
christopherhein: Yeah, this might have been a miss step cause you are right, this means it's going to mask the problem. Basically `defer` and `os.Exit` don't work together so when we're exiting our binary it's actually not writing the logs, this moved it out of a func with `os.Exit` but I'm realizing that it's not going to actually defer anything now. Updating.

christopherhein: Yeah, I'm just going to get rid of `klog` in the capn packages cause we don't even use it outside of `main.go` and we've been using controller-runtime's logger so it's duplicative.

christopherhein: Actually we don't even use this log package, we're using klogr for the controller runtime logger so this isn't even used, deleting all of it :)

PR #273, opened 2022-05-11
✨ Add SuperClusterLabelling experimental feature gate

### LINKED ISSUE #272: Add a label to all namespaces created by virtualcluster controller
**User Story**
As an operator I would like to have more control over resources listed by controllers to reduce the overall load to the API

**Detailed Description**
VirtualCluster controller and in particular namespace syncer garbage collector list all resources in the cluster and only then filters them by virtualcluster annotations (like tenancy.x-k8s.io/vcname) https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/1e987ad6c3c6ef0099e898b1b6589975212523b9/virtualcluster/pkg/syncer/resources/namespace/checker.go#L90

We should consider adding a label to these namespaces, like `tenancy.x-k8s.io/controlled: true` and use this label in the listing mentioned.

Consider propagating this label to all resources created by the virtual cluster controllers in the follow-ups

/kind feature

m-messiah:
This PR adds a new feature gate `SuperClusterLabelling` which enables adding labels to the namespaces created by syncer.
This behaviour could be extended for any other resources, but starts from namespaces as the particular case. 

fixes #272, requires a follow-up with enabling labels for filtering too: https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/1e987ad6c3c6ef0099e898b1b6589975212523b9/virtualcluster/pkg/syncer/resources/namespace/checker.go#L90

The changes are intentionally split by different PRs to allow operator to ensure all namespace are already labelled before reducing the scope of syncer/resources/namespace/checker: garbage collector

### DIFF
--- a/controlplane/nested/certificate/helpers.go
+++ b/controlplane/nested/certificate/helpers.go
@@ -76,7 +76,7 @@ func NewAPIServerCrtAndKey(ca *KeyPair, clusterName, clusterDomainArg, apiserver
 // NewAPIServerKubeletClientCertAndKey creates certificate for the apiservers to connect to the
 // kubelets securely, signed by the ca.
 // A hack to use namespace as CN because vn-agent is using this CN to figure out the namespace in
-// super cluster for kubectl exec/log/port forward
+// super cluster for kubectl exec/log/port forward.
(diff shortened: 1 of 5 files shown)


christopherhein (approved):
/lgtm
/approve

PR #274, opened 2022-05-12
✨ Add SuperClusterLabelFilter experimental feature gate

m-messiah:
The PR is a follow-up feature gate to SuperClusterLabelling https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/273.
It adds new feature gate SuperClusterLabelFilter with namespace checker implementation to allow use label filter requesting super cluster namespaces, instead of doing the full list. It should reduce load to super cluster control plane in multi-platform super clusters (i.e. when super cluster itself is multi-tenant in addition to virtual clusters).

### DIFF
--- a/virtualcluster/pkg/syncer/conversion/helper.go
+++ b/virtualcluster/pkg/syncer/conversion/helper.go
@@ -197,6 +197,15 @@ func (c *objectConversion) CleanOpaqueKeys(vc *v1alpha1.VirtualCluster, keyMap m
 	}
 }
 
+func WithSuperClusterLabels(labels map[string]string) map[string]string {
+	if labels == nil {
+		labels = make(map[string]string)
+	}
+
+	labels[constants.LabelControlled] = "true"
+	return labels
+}
+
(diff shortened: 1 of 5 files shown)


virtualcluster/pkg/syncer/conversion/helper.go
```diff
@@ -197,6 +197,15 @@ func (c *objectConversion) CleanOpaqueKeys(vc *v1alpha1.VirtualCluster, keyMap m
 	}
 }
 
+func WithSuperClusterLabels(labels map[string]string) map[string]string {
```
m-messiah, line 200: I made it `With..` oriented to allow simple chaining, but wondering, if nil check is needed there and should we just use `ApplyLabels(labels)` instead, rather than chaining map. Any comments?

christopherhein:
@m-messiah I wonder since the conversion happens for all objects if it would make sense to make this a feature for all the resource syncers.

christopherhein (approved):
/lgtm
/approve

PR #279, opened 2022-06-13
✨ Use SuperClusterLabelling feature gate for all objects if enabled

m-messiah:
SuperClusterLabelling was initially created for namespaces, but there are a lot of other objects and controllers created in virtualclusters. The PR adds labeling for all objects too, which could help in the future to use SuperClusterLabelFilter feature gate for any resource. (or even do something like `kubectl get pod -A -l tenancy.x-k8s.io/controlled=true` to find only tenant pods in the Kubernetes cluster).


christopherhein (approved):
/lgtm
/approve

PR #280, opened 2022-06-16
✨ Use SuperClusterLabelFilter feature gate for all objects if enabled

m-messiah:
#279 introduced labelling for all objects, so this PR adds filtering to all checkers where it is possible.

There are couple checkers (priority class, storage class, persistent volume) that are used for back populating only and have nothing to do with virtual clusters initially, so they still have labels.Everything() selector. All other checkers are used for synchronisation and garbage collecting, so labelling should help to reduce control-plane operations and overall controller memory consumption in clusters, where super-cluster is used outside of nested scope.


christopherhein (commented):
/lgtm

I would love to get @Fei-Guo 👁️ on this as well.

Fei-Guo:
Thanks for working on this.
/lgtm
/approve

PR #284, opened 2022-07-07
⚠️ Remove the usage of non-inclusive language

m-messiah:

The PR removes the usage of non-inclusive language and terminology.

Most of it is internal, but there are some major changes:

- `--master-prov` flag is renamed to `--provisioner`. The previous flag is still exists as deprecated to avoid breaking.
- Metric resources and event reasons like `DeletedOrphanSuper[Master]ConfigMaps` are renamed to `DeletedOrphanSuper[ControlPlane]ConfigMaps`
- tenant apiserver flag `--endpoint-reconciler-type=master-count` is removed as it should use leases by default since k8s 1.13.

There are some `master` words remains - in syncer and scheduler flags (probably we need to do the same deprecateion process as for `--master-prov` flag.
There is also the default rbac group `system:masters` that is not changed.


Fei-Guo:
Awesome! Thanks for the clean ups. 

/lgtm
/approve

PR #286, opened 2022-07-13
✨ Enable golangci-lint plugins for virtual cluster project 

m-messiah:
**What this PR does / why we need it**:
The PR https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/285 enabled some linters, but there are many to go. 
The PR enables them as well and fixes related problems

PR #285, opened 2022-07-07
✨ Add golangci-lint for virtual cluster project

m-messiah:
**What this PR does / why we need it**:
Previously virtualcluster object has no linting and had some static golang problems or readability issues.
The PR is intended to enable linting for the directory.
As the initial run showed more than 2500 issues the PR consists of many commits for enabling golangci-lint checkers one by one.

PR #287, opened 2022-07-15
✨ Enable all golangci-lint plugins for virtual cluster project

m-messiah:

This is the last part from three-step golangci-lint enable, started in https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/285 and https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/286.

This part is focused on the gosimple, gosec, gocyclo and revive checkers, bringing little more readability and memory usage efficiency.

## Notes

1. The `revive` plugin has two exclusions, because otherwise it would require to write too much comments, which could not make a lot of sense:

```yaml
# EXC0012 revive: Annoying issue about not having a comment. The rare codebase has such comments
- exported (.+) should have comment( \(or a comment on this block\))? or be unexported
# EXC0013 revive: Annoying issue about not having a comment. The rare codebase has such comments
- package comment should be of the form "(.+)...
``` 

## Notable changes

1. kubectl-vc kubeconfig creation mode changed to 0600 instead of 0644. There is no sense to make config readable for other users, but if you have any objections - I could revert it and disable in `gosec` linter.
3. Minimum TLS version for TLSConfig is pinned to TLS1.2. I can't imaging anybody to use older and less secure, but it is still one of major changes
4. provisioner.ProvisionerAliyun is renamed to provisioner.Aliyun due to unnecessary duplication of the word _provisioner_
5. provisioner.ProvisionerNative is renamed to provisioner.Native due to unnecessary duplication of the word _provisioner_
6. validationplugin.ValidationPluginInterface is renamed to validationplugin.Interface due to unnecessary duplication of the word _validationplugin_
7. fairqueue.FifoQueue is exported type now


christopherhein (approved):
/lgtm

/cc @Fei-Guo do you have an opportunity to give this a scan as well? I can't see anything too concerning.

virtualcluster/pkg/syncer/syncer.go
```diff
@@ -233,25 +233,6 @@ func (s *Syncer) Run(stopChan <-chan struct{}) {
 		}
 	}()
 	go wait.Until(s.healthPatrol, 1*time.Minute, stopChan)
-	go func() {
```
Fei-Guo, line 236: @m-messiah  Why do you remove this go func?

m-messiah: @Fei-Guo thanks for catching this up. I made a revert PR for this part of code: https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/288 

I retried it and it looks like when I hit `remove redundant return` in Goland - it removes this goroutine too..

christopherhein: Thanks for catching that Fei!

PR #289, opened 2022-07-18
🐛 Add virtualcluster tests to CI

m-messiah:

virtual cluster tests were disabled in https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/267 and we had no CI tests since that. 
The PR enables them back.

The envtest is failing to start when `--race` flag is set, so I removed it.

Also, SHA1 is proved to be weak and golang tls prevented these certificate from being used https://github.com/golang/go/issues/41682. The PR regenerates test certificates with sha256

Also fix ginkgo warning:
> You are passing a Done channel to a test node to test asynchronous behavior.  This is deprecated in Ginkgo V2.  Your test will run synchronously and the timeout will be ignored.
> Learn more at: https://github.com/onsi/ginkgo/blob/v2/docs/MIGRATING_TO_V2.md#removed-async-testing


christopherhein (approved):
/lgtm
/approve 

Thank you for fixing all of this up!

PR #290, opened 2022-07-19
✨ Control Plane Upgrades

m-messiah:
The PR adds the featureGate `ClusterVersionPartialUpgrade` to allow native provide to reconcile currently running virtual clusters to re-apply clusterversion specs to control plane.

### Changes to the default behaviour or code
The PR introduces changes to the current native provisioner (Aliyun just returns `not implemented` for the feature):
1. Use serverSide Apply instead of Create means we can reuse methods to re-apply state, instead of breaking with AlreadyExists conflicts. 
2. Kubernetes client requests are finally share the current context instead of TODO().
3. ClusterVersion must contain apiVersion and kind for each Service and StatefulSet defined (it exists in tests, but could be omitted in real world. Using Patch require them to be set)

### Features of the featureGate
1. The change requires control plane statefulsets to have RollingUpdateStatefulSetStrategyType instead of OnDelete to allow rollouts after upgrade.
2. APIServer and ControllerManager do not watch certificates changes on disk, so the statefulsets are annotated to force pod recreation if the certificates changed. In real, the certificates are refreshed in each upgrade, while RootCA remains untouched.
3. ETCD is already reloads certificates from the disk, so it does not need to be annotated
4. Two metrics introduced:
```
# HELP clusters_upgrade_seconds Duration of cluster upgrade by reconciler in featuregate.ClusterVersionPartialUpgrade
# TYPE clusters_upgrade_seconds histogram
clusters_upgrade_seconds_bucket{cluster_version="v1.19-test",resource_version="857463552",le="10"} 1
clusters_upgrade_seconds_bucket{cluster_version="v1.19-test",resource_version="857463552",le="+Inf"} 1
clusters_upgrade_seconds_sum{cluster_version="v1.19-test",resource_version="857463552"} 7.5169857780000005
clusters_upgrade_seconds_count{cluster_version="v1.19-test",resource_version="857463552"} 1
# HELP clusters_upgraded Amount of clusters upgraded by reconciler in featuregate.ClusterVersionPartialUpgrade
# TYPE clusters_upgraded counter
clusters_upgraded{cluster_version="v1.19-test",resource_version="857463552"} 1
```

### DIFF
--- a/virtualcluster/pkg/controller/controllers/metrics.go
+++ b/virtualcluster/pkg/controller/controllers/metrics.go
@@ -2,9 +2,6 @@ package controllers
 
 import (
 	"github.com/prometheus/client_golang/prometheus"
-	"sigs.k8s.io/controller-runtime/pkg/metrics"
-
-	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/featuregate"
 )
 
 var (
@@ -23,10 +20,3 @@ var (
 		[]string{"cluster_version", "resource_version"},
 	)
 )
-
-func init() {
-	// Expose featuregate.ClusterVersionApplyCurrentState metrics only if it enabled
-	if featuregate.DefaultFeatureGate.Enabled(featuregate.ClusterVersionApplyCurrentState) {
-		metrics.Registry.MustRegister(clustersUpdatedCounter, clustersUpdateSeconds)
-	}
-}
--- a/virtualcluster/pkg/controller/controllers/metrics.go
+++ b/virtualcluster/pkg/controller/controllers/metrics.go
@@ -14,8 +14,9 @@ var (
 	)
 	clustersUpdateSeconds = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name: "clusters_update_seconds",
-			Help: "Duration of cluster upgrade by reconciler in featuregate.ClusterVersionApplyCurrentState",
+			Name:    "clusters_update_seconds",
+			Help:    "Duration of cluster upgrade by reconciler in featuregate.ClusterVersionApplyCurrentState",
+			Buckets: []float64{.1, .5, 1, 5, 10, 20, 30, 60, 90, 120, 300, 600, 900},
 		},
 		[]string{"cluster_version", "resource_version"},
 	)
--- a/virtualcluster/pkg/controller/controllers/metrics.go
+++ b/virtualcluster/pkg/controller/controllers/metrics.go
@@ -5,17 +5,24 @@ import (
 )
 
 var (
-	clustersUpdatedCounter = prometheus.NewCounterVec(
+	clustersUpgradedCounter = prometheus.NewCounterVec(
 		prometheus.CounterOpts{
-			Name: "clusters_updated",
-			Help: "Amount of clusters upgraded by reconciler in featuregate.ClusterVersionApplyCurrentState",
+			Name: "clusters_upgraded",
+			Help: "Amount of clusters upgraded by reconciler in featuregate.ClusterVersionPartialUpgrade",
 		},
 		[]string{"cluster_version", "resource_version"},
 	)
-	clustersUpdateSeconds = prometheus.NewHistogramVec(
+	clustersUpgradeFailedCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "clusters_upgrade_failed",
+			Help: "Amount of clusters failed to upgrade by reconciler in featuregate.ClusterVersionPartialUpgrade",
+		},
+		[]string{"cluster_version", "resource_version"},
+	)
+	clustersUpgradeSeconds = prometheus.NewHistogramVec(
 		prometheus.HistogramOpts{
-			Name:    "clusters_update_seconds",
-			Help:    "Duration of cluster upgrade by reconciler in featuregate.ClusterVersionApplyCurrentState",
+			Name:    "clusters_upgrade_seconds",
+			Help:    "Duration of cluster upgrade by reconciler in featuregate.ClusterVersionPartialUpgrade",
 			Buckets: []float64{.1, .5, 1, 5, 10, 20, 30, 60, 90, 120, 300, 600, 900},
 		},
 		[]string{"cluster_version", "resource_version"},
--- a/virtualcluster/pkg/controller/controllers/metrics.go
+++ b/virtualcluster/pkg/controller/controllers/metrics.go
@@ -1,0 +1,32 @@
+package controllers
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"sigs.k8s.io/controller-runtime/pkg/metrics"
+
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/featuregate"
+)
+
+var (
+	clustersUpdatedCounter = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "clusters_updated",
+			Help: "Amount of clusters upgraded by reconciler in featuregate.ClusterVersionApplyCurrentState",
+		},
+		[]string{"cluster_version", "resource_version"},
+	)
+	clustersUpdateSeconds = prometheus.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name: "clusters_update_seconds",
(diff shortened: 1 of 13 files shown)


m-messiah:
> I am trying to understand this feature. In the original design, we intend to make clusterversionName field immutable in VC CRD. Are you targeting a user case where user changes the clusterversionName? or you want to handle cases where the clusterversion CR itself is updated and you want to populate the change to all the running VCs?

The second. I want (as a cluster operator) to release new minor versions of clusterversion by adding/updating apiserver flags, updating minor versions of apiserver or controller-manager or manipulate the service params according to the normal process of software upgrades or sudden security patches. So, in this feature we trying to implement a way to populate the current clusterVersion (CR) state to virtual clusters that use that version and agreed to be upgraded (by setting a label).

In foreseeable future it is intended to be used for minor upgrades only, keeping different clusterVersion CRs for different major k8s versions.

Just now, it is already useful to change log verbosity or enable/disable some authorisation params, which anyway "are managed from outside"

Fei-Guo:
> > I am trying to understand this feature. In the original design, we intend to make clusterversionName field immutable in VC CRD. Are you targeting a user case where user changes the clusterversionName? or you want to handle cases where the clusterversion CR itself is updated and you want to populate the change to all the running VCs?
> 
> The second. I want (as a cluster operator) to release new minor versions of clusterversion by adding/updating apiserver flags, updating minor versions of apiserver or controller-manager or manipulate the service params according to the normal process of software upgrades or sudden security patches. So, in this feature we trying to implement a way to populate the current clusterVersion (CR) state to virtual clusters that use that version and agreed to be upgraded (by setting a label).
> 
> In foreseeable future it is intended to be used for minor upgrades only, keeping different clusterVersion CRs for different major k8s versions.
> 
> Just now, it is already useful to change log verbosity or enable/disable some authorisation params, which anyway "are managed from outside"

Thanks. Just FYI, we are trying to resolve the problems you mentioned using CAPN. We didn't handle vc update in the native provisioner because it was designed for PoC purpose, and we don't recommend using it in production. We want to make it simple. As you may see, the etcd Pod does not even use a persistent storage there. I am glad you are looking into the update problem, but I would suggest enhancing CPAN provider.

christopherhein:
@Fei-Guo Per our chats in Slack, this is more of stop-gap while we haven't been able to move over to CAPN, I think it's worthwhile in the interim to still move this forward since it doesn't hurt anything necessarily. We've been able to overcome a lot of those challenges by maintaining our own custom ClusterVersions with persistence for example and we're trying to keep etcd upgrades out of the scope for the vc-manager. 

Do you still object to this development?

Fei-Guo:
> @Fei-Guo Per our chats in Slack, this is more of stop-gap while we haven't been able to move over to CAPN, I think it's worthwhile in the interim to still move this forward since it doesn't hurt anything necessarily. We've been able to overcome a lot of those challenges by maintaining our own custom ClusterVersions with persistence for example and we're trying to keep etcd upgrades out of the scope for the vc-manager.
> 
> Do you still object to this development?

Thanks for the clarification. I have no problem with enhancing the clusterversion based vc lifecycle management, but I hope the feature scope should be very clear. As we discussed offline, we may name the feature more specifically, something like "ClusterVersionPartialUpdate" or better name.

virtualcluster/pkg/controller/controllers/provisioner/provisioner_native.go
```diff
@@ -240,62 +295,85 @@ func (mpn *Native) createPKISecrets(caGroup *vcpki.ClusterCAGroup, namespace str
...
-	rootRsaKey, ok := rootKey.(*rsa.PrivateKey)
-	if !ok {
-		return errors.New("fail to assert rsa PrivateKey")
-	}
+	var rootCAPair *vcpki.CrtKeyPair
+
+	// reuse rootCa if it is present
+	rootCaSecret := &corev1.Secret{}
+	err := mpn.Get(ctx, client.ObjectKey{Name: secret.RootCASecretName, Namespace: vc.Status.ClusterNamespace}, rootCaSecret)
+	if err == nil {
```
christopherhein, line 322: Should we change this to add `&& errors.IsNotFound(err)`? I'm worried that if we have a network blip or something else is causing the `Get` to error then we'd recreate this object and it would end up in a not clean state.

christopherhein (changes requested):
This is looking really good, a couple comments with the main theme being remove the etcd upgrades. Thanks!

virtualcluster/pkg/controller/controllers/virtualcluster_controller.go
```diff
@@ -163,6 +175,42 @@ func (r *ReconcileVirtualCluster) Reconcile(ctx context.Context, request reconci
...
+		err = r.Provisioner.UpgradeVirtualCluster(ctx, vc)
+		clustersUpdateSeconds.WithLabelValues(vc.Spec.ClusterVersionName, vc.Labels[constants.LabelClusterVersionApplied]).Observe(time.Since(upgradeStartTimestamp).Seconds())
+		if err != nil {
+			r.Log.Error(err, "fail to upgrade virtualcluster", "vc", vc.GetName())
+			kubeutil.SetVCStatus(vc, tenancyv1alpha1.ClusterRunning, fmt.Sprintf("fail to upgrade: %s", err), "TenantControlPlaneUpgradeFailed")
+		} else {
+			r.Log.Info("upgrade finished", "vc", vc.GetName())
+			kubeutil.SetVCStatus(vc, tenancyv1alpha1.ClusterRunning, "tenant control plane is upgraded", "TenantControlPlaneUpgradeCompleted")
+			clustersUpdatedCounter.WithLabelValues(vc.Spec.ClusterVersionName, vc.Labels[constants.LabelClusterVersionApplied]).Inc()
+		}
```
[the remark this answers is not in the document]
m-messiah: no, we have no return in Failed branch, because we still want to set cluster status and remove ready-for-upgrade label later

virtualcluster/pkg/controller/controllers/provisioner/provisioner_native.go
m-messiah: correct

virtualcluster/pkg/controller/controllers/provisioner/provisioner_native.go
```diff
@@ -61,60 +67,93 @@ func NewProvisionerNative(mgr manager.Manager, log logr.Logger, provisionerTimeo
...
 	if err != nil {
 		return err
 	}
 
 	// 3. deploy etcd
-	err = mpn.deployComponent(vc, cv.Spec.ETCD)
+	err = mpn.deployComponent(ctx, vc, cv.Spec.ETCD, clusterCAGroup)
 	if err != nil {
 		return err
 	}
```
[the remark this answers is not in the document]
m-messiah: I made an ETCD nil assignment as a workaround for now, let's do upgrade code later, when we decide it would be needed.

virtualcluster/pkg/controller/controllers/virtualcluster_controller.go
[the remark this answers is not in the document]
christopherhein: 👍

virtualcluster/pkg/controller/controllers/provisioner/provisioner_native.go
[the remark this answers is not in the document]
christopherhein: Nice, I like that!

christopherhein (approved):
/lgtm
/approve

Amazing work! 🎉

virtualcluster/pkg/controller/controllers/metrics.go
```diff
@@ -0,0 +1,30 @@
+package controllers
```
Fei-Guo, line 1: Please add file headers for all newly created files.

Fei-Guo:
Also, please add a document to describe the workflow in tutorial. Based on my understanding, the workflow of triggering update is 
1) Change the clusterversion cr
2)  Add a label in the VC cr
which is not obvious for many users.

virtualcluster/pkg/controller/controllers/metrics.go
christopherhein: Specifically the license contents @m-messiah , thanks!

m-messiah: will be added in https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/292 (with linter)

Issue #282, opened 2022-07-07
✨ Projected ServiceAccount Support


christopherhein:
**User Story**

As a user I would like to use ServiceAccount Projection from a VirtualCluster so that I can expose my cluster as an Identity provider and issue certs to tools like vault.

**Detailed Description**

Upstream Kubernetes added support for ServiceAccount projection but it requires that the kubelet make requests on behalf of a workload to grant it tokens with specific audiences and expirations this is difficult to do with shared kubelets in VC. It would be nice if we could figure a way to support this.

Docs: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#service-account-token-volume-projection

**Anything else you would like to add:**

You can make these with a TokenRequest definition and using curl against the token subresource for ServiceAccounts like so:

```
curl -X "POST" {API}/api/v1/namespaces/{namespace}/serviceaccounts/{serviceaccount-name}/token -H 'Authorization: Bearer {token}' -H 'Content-Type: application/json; charset=utf-8' -d $'{}'
```

the Body of this usually accepts this object - https://github.com/kubernetes/api/blob/v0.20.1/authentication/v1/types.go#L131

For example:
```
{
  "apiVersion": "authentication.k8s.io/v1",
  "kind": "TokenRequest",
  "metadata": {
    "name": "{name}"
  },
  "spec": {
    "audiences": [
      "vault"
    ],
    "expirationSeconds": 7200,
    "boundObjectRef": {
      "kind": "Pod",
      "apiVersion": "v1",
      "name": "{name}",
      "uid": "{uid}"
    }
  }
}
```

**Ideas**

My initial thought process is adding a step in the pod syncer, where it currently checks for ServiceAccount tokens being created before creating the pod that it would call out to a "projected serviceaccount syncer" to make this request against the tenant control plane then stores the contents in a secret only in the super cluster then we mutate the pod spec to change from projected to a secret source and mount like a normal secret. We'd then need something that was checking validity of these tokens, the "projected serviceaccount syncer" for example, and when it got close to expiring we redo the call and update the token causing the secret mount to update as well.

Alternatively, we could modify the Kubelet to be "tenant aware" but I imagine this would become a massive effort.

/kind feature

Fei-Guo:
[...]

PR #295, opened 2022-07-22
🐛 Resource already exists and the UID is different should not requeue

### LINKED ISSUE #293: Resource already exists and the UID is different should not requeue
**What would you like to be added?**
When the DWS synchronization resource is encountered but its delegated object UID is different in the `reconcileXXXCreate` process, it should not re-enter the queue, because the event will also not be processed in the subsequent process.

Currently:
```go
	if errors.IsAlreadyExists(err) {
		if pPod.Annotations[constants.LabelUID] == requestUID {
			klog.Infof("pod %s/%s of cluster %s already exist in super master", targetNamespace, pPod.Name, clusterName)
			return nil
		} else {
			return fmt.Errorf("pPod %s/%s exists but the UID is different from tenant master.", targetNamespace, pPod.Name)
		}
	}
```

Expect:
```go
	if errors.IsAlreadyExists(err) {
		if pPod.Annotations[constants.LabelUID] == requestUID {
			klog.Infof("pod %s/%s of cluster %s already exist in super master", targetNamespace, pPod.Name, clusterName)
			return nil
		} else {
			klog.Errorf("pPod %s/%s exists but the UID is different from tenant master.", targetNamespace, pPod.Name)
			return nil
		}
	}
```

**Why is this needed?**
If the queue is not dequeued so early, the abnormal tasks will continue to accumulate until they exceed the maximum number of retries `MaxReconcileRetryAttempts` will be discarded.

/kind bug

wondywang:
If the queue is not dequeued so early, the abnormal tasks will continue to accumulate until they exceed the maximum number of retries MaxReconcileRetryAttempts will be discarded.

So, when the DWS synchronization resource is encountered but its delegated object UID is different in the reconcileXXXCreate process, it should not re-enter the queue, because the event will also not be processed in the subsequent process.

Fixes #293

### DIFF
--- a/virtualcluster/pkg/syncer/resources/configmap/dws.go
+++ b/virtualcluster/pkg/syncer/resources/configmap/dws.go
@@ -96,9 +96,10 @@ func (c *controller) reconcileConfigMapCreate(clusterName, targetNamespace, requ
 	if apierrors.IsAlreadyExists(err) {
 		if pConfigMap.Annotations[constants.LabelUID] == requestUID {
 			klog.Infof("configmap %s/%s of cluster %s already exist in super control plane", targetNamespace, configMap.Name, clusterName)
-			return nil
+		} else {
+			klog.Errorf("pConfigMap %s/%s exists but its delegated object UID is different", targetNamespace, pConfigMap.Name)
 		}
-		return fmt.Errorf("pConfigMap %s/%s exists but its delegated object UID is different", targetNamespace, pConfigMap.Name)
+		return nil
 	}
 	return err
 }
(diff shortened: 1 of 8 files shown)


Fei-Guo:
@wondywang I think it is ok to have retry if UID mismatch, otherwise if user deletes->creates the resource with the same name in VC and the order happens is create->delete in the super cluster, the new resource may take 1 min to be created in super after the patroller clear the stale one in the super cluster. 

Also, this change does not resolve the problem that the patroller will delete the Kube-root-ca cfgmap in the super cluster periodically (although the cfgmap will be created automatically again).  Note that Issue #293 and #282 are related since the Kube-root-ca is part of the projected service account. So here is my suggestion:

1) Add a feature gate for supporting projected service account;
2) As a temporal solution, in cfgmap DWS, add a whitelist to skip the cfgmap with name "Kube-root-ca" (using the feature gate to provide backward compatibility since user may create a cfgmap with this name before 1.20);
3) Also change the cfgmap patroller to skip the "Kube-root-ca" cfgmap;

Later, we can add code to fully support the projected service account.
What do you think?

wondywang:
> 

Thanks @Fei-Guo , now I know I the retry is neede. Meanwhile, a whitelist may be a good temporal solution for "kube-root-ca". Then, in the end, we have to plan how to support the projected service account.

So, I will close this PR.

PR #299, opened 2022-08-25
🌱 Update CRD Clients to v1

christopherhein:
Updates the CRD types to v1 from v1beta1, v1beta1 was deprecated and when using the crd syncer it will fail to boot on anything above v1.20.

### DIFF
--- a/virtualcluster/pkg/syncer/conversion/equality.go
+++ b/virtualcluster/pkg/syncer/conversion/equality.go
@@ -19,12 +19,11 @@ package conversion
 import (
 	"strings"
 
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
-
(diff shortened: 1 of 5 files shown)


virtualcluster/pkg/syncer/conversion/equality.go
```diff
@@ -19,7 +19,7 @@ package conversion
 import (
 	"strings"
 
-	"k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1beta1"
+	apiextensionsv1 "k8s.io/apiextensions-apiserver/pkg/apis/apiextensions/v1"
```
m-messiah, line 23: you can remove line there, and re-sort imports.

christopherhein: Done 👍

m-messiah (approved):

PR #300, opened 2022-08-30
🌱 Remove excess NodeInfo.OperatingSystem assign

m-messiah:
**What this PR does / why we need it**: The PR removes confusing excess n.Status.NodeInfo.OperatingSystem assign to Linux, which is not used anywhere and is rewritten in a couple of lines later https://github.com/m-messiah/cluster-api-provider-nested/blob/02b288a700d6ef6fd1776f4d63e3e4a7e0ca317c/virtualcluster/pkg/syncer/vnode/vnode.go#L100


christopherhein (approved):
/lgtm
/approve

PR #301, opened 2022-09-01
🌱 Update samples with k8s 1.22 and apiVersions

m-messiah:
The PR updates sample clusterversions to use k8s 1.22 and adds missing apiVersion/kind fields, required for provisioner.

Note, etcd 3.4 uses ETCDCTL_API=3 by default, so we don't need to prepend this anymore


atosatto (approved):

christopherhein (approved):
/lgtm
/approve

Thanks @m-messiah for doing this!

PR #302, opened 2022-09-08
🌱 Add etcd.namespace to the etcd certificate

m-messiah:
Currently, etcd certificate is valid for `etcd-0.etcd, etcd-1.etcd, etcd-2.etcd, etcd` names only. That prevents us to connect to etcd from controllers to do health checks, upgrade maintenance or backups without using InsecureSkipVerify.
The PR adds virtualcluster namespace to every etcd domain, so they would be accessible


christopherhein (approved):
/lgtm
/approve

PR #305, opened 2022-09-14
✨ Sync node taints and labels

m-messiah:

***Context***:

The main reason for these changes is to allow VirtualCluster users to use pod.Spec.NodeSelector in their pods and choose nodes they want to be assigned. This would be extended after to auto-apply this nodeSelector for all pods in syncer if we want to have VirtualCluster fully isolated on the subset of nodes.

***Technical description***:

The PR adds sync of node Taints from super cluster to VirtualCluster to allow VirtualCluster pods to have tolerations.
The PR also extends the list of node labels to be synced to VirtualCluster, providing a new flag `--extra-node-labels=<stringslice>` to allow an operator to set the list of labels that need to be accessible by VirtualCluster users and used by them as node selectors.

***Tests***:

The change was tested by passing `--extra-node-labels=node-selectors/default,test-label` to the syncer, creating VirtualCluster with a pod and checking node labels in node from inside the VirtualCluster. 
I also edited super-cluster node labels and it was successfully synced to the VirtualCluster.

### DIFF
--- a/virtualcluster/pkg/syncer/vnode/provider/provider.go
+++ b/virtualcluster/pkg/syncer/vnode/provider/provider.go
@@ -18,6 +18,7 @@ package provider
 
 import (
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/constants"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/featuregate"
@@ -29,6 +30,7 @@ type VirtualNodeProvider interface {
 	GetNodeDaemonEndpoints(node *corev1.Node) (corev1.NodeDaemonEndpoints, error)
 	GetNodeAddress(node *corev1.Node) ([]corev1.NodeAddress, error)
 	GetLabelsToSync() map[string]struct{}
+	GetTaintsToSync() map[string]struct{}
 }
 
 // GetNodeLabels is used to sync allowed node labels to vNode
@@ -48,4 +50,31 @@ func GetNodeLabels(p VirtualNodeProvider, node *corev1.Node) map[string]string {
 		}
 	}
 	return labels
+}
+
+// GetNodeTaints is used to convert pNode taints to vNode (adding corev1.TaintNodeUnschedulable)
+func GetNodeTaints(p VirtualNodeProvider, node *corev1.Node, now metav1.Time) (taints []corev1.Taint) {
+	nodeTaints := node.Spec.Taints
+	newTaint := corev1.Taint{
+		Key:       corev1.TaintNodeUnschedulable,
+		Effect:    corev1.TaintEffectNoSchedule,
+		TimeAdded: &now,
+	}
+	taintsToSync := p.GetTaintsToSync()
+	updated := false
+	for i := range nodeTaints {
+		if newTaint.MatchTaint(&nodeTaints[i]) {
+			nodeTaints[i].TimeAdded = newTaint.TimeAdded
+			updated = true
+			continue
+		}
+		if _, found := taintsToSync[nodeTaints[i].Key]; !found {
+			continue
+		}
+		taints = append(taints, nodeTaints[i])
+	}
+	if !updated {
+		taints = append(taints, newTaint)
+	}
+	return taints
 }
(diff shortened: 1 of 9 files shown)


Fei-Guo:
VC starts with no nodes in the cluster. How will this PR help with the creation of the first Pod?

m-messiah:
> VC starts with no nodes in the cluster. How will this PR help with the creation of the first Pod?

This PR has nothing to do with pods, and yes, for the first pod you would need to know what selectors are available in the cluster. But this would be a different PR and solution to share the knowledge about nodeselectors. 
Right now, the first pod could request everything it wants, and the vNode reconciler will bring the node to the VirtualCluster. 
So this PR just allows labels and taints to be visible in vNode and does nothing with scheduling

virtualcluster/pkg/syncer/vnode/vnode.go
```diff
@@ -104,7 +85,31 @@ func NewVirtualNode(provider provider.VirtualNodeProvider, node *corev1.Node) (v
 	return n, nil
 }
 
-var wellKnownNodeLabelsMap = map[string]struct{}{
+// BuildVNodeTaints is used to convert pNode taints to vNode (adding corev1.TaintNodeUnschedulable)
+func BuildVNodeTaints(node *corev1.Node, now metav1.Time) (vnodeTaints []corev1.Taint) {
+	nodeTaints := node.Spec.Taints
```
Fei-Guo, line 90: I am not sure if blindly exposing all taints of super node to virtual cluster has any side effects about breaking isolation requirements. Hence, I would suggest having this capability guarded by a feature flag.

virtualcluster/pkg/syncer/vnode/provider/provider.go
```diff
@@ -18,10 +18,34 @@ package provider
...
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.SuperClusterPooling) {
+		labels[constants.LabelSuperClusterID] = utilconstants.SuperClusterID
+	}
+
+	labelsToSync := p.GetLabelsToSync()
+	for k, v := range node.GetLabels() {
+		if _, found := labelsToSync[k]; found {
+			labels[k] = v
+		}
+	}
```
christopherhein, line 49: Does this need to go through the same filtering we do on https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/pkg/syncer/conversion/equality.go#L117-L162 so that we don't sync labels labels in the VirtualCluster CR?

virtualcluster/pkg/syncer/vnode/vnode.go
christopherhein: Would a new field in the VirtualCluster CR that allows a (or excludes) a list of taints like `opaqueMetaPrefixes` & `transparentMetaPrefixes` be useful?

Fei-Guo: As I asked, it depends on how you plan to use the taints. How would the cluster admin know the taints to be included/excluded. For this PR, node taints are blindly copied. Also remember that this field should be fixed during VC lifecycle, is it applicable for taints.

virtualcluster/pkg/syncer/vnode/provider/provider.go
m-messiah: The list is pretty straightforward as comes from the command line flags, so the operator should explicitly add the label to be synced.
On the other hand, the CheckDWObjectMetaEquality method is not used for vNode and pNode comparison, so there are no conflicts.

virtualcluster/pkg/syncer/vnode/vnode.go
m-messiah: > Would a new field in the VirtualCluster CR that allows a (or excludes) a list of taints like `opaqueMetaPrefixes` & `transparentMetaPrefixes` be useful?

I could do the same flag as for labels, but for the list of Taint keys. Initially, I did not think we could expose any isolation or privacy by just seeing the list of taints (it is still the pod syncer and scheduler problems to allow or forbid tolerations  and nodeSelector usage in pods), but yes, let's do it as an another allowlist

m-messiah: Added a flag, please take a look

m-messiah: @Fei-Guo do you think the PR is good now?

PR #307, opened 2022-09-18
🐛 Add Annotation UID of pPV to be synced to vPV

wondywang:
What type of PR is this?
/kind bug

What this PR does / why we need it:
Add Annotation UID of pPV to be synced to vPV, which is used to identify for consistency check.

Which issue(s) this PR fixes:
Fixes #


wondywang:
PTAL @christopherhein

PR #309, opened 2022-09-22
🐛 Fix cancelFunc on cluster init and fix node Patch

m-messiah:
The PR fixes two bugs found in syncer:
1. cancelFunc is not initialised in the cluster before runCluster method, which sometimes leads to nil-pointer panic when we want to remove virtualcluster with the broken control-plane. I initialize it to a noop func to avoid panic. (I found the same broken cluster, removed it and there are no panic in syncer then)
2. Node Patch for Status and Spec leads to conflicts as patch status implicitly updates ObjectMeta too (labels and ResourceVersion), and spec update then fails with "object is modified". The PR ensures Status patch is for Status only and the Spec + Label patch is for Spec and Labels only.

### DIFF
--- a/virtualcluster/pkg/syncer/vnode/vnode.go
+++ b/virtualcluster/pkg/syncer/vnode/vnode.go
@@ -141,12 +141,13 @@ func nodeConditions() []corev1.NodeCondition {
 }
 
 func UpdateNode(client v1core.NodeInterface, node, newNode *corev1.Node) error {
-	updatedNode, _, err := patchNodeStatus(client, types.NodeName(node.Name), node, newNode)
-	if err != nil {
-		return err
-	}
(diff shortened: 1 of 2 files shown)


christopherhein (approved):
/lgtm
/approve

I'm okay with this, I do think we need to start thinking about how the conversations package & the vnode package can combined over time, it seems like we should have a lot of the same filtering that is happening in the conversion packages. EG there are some builtins we have to make sure resourceVersions and generation are removed.

https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/e489733330e0e7cb2ef9b679bb3d5acb3ffbc407/virtualcluster/pkg/syncer/conversion/helper.go#L268-L278

PR #310, opened 2022-09-23
Fix Pod Checker occasionally deletes vPods unexpectedly

### LINKED ISSUE #308: Pod Checker occasionally deletes vPods unexpectedly

When there are a large number of VCs in the cluster, vPods are unexpectedly deleted by the Pod Checker accidentally. The logic that triggers deletion is in https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/pkg/syncer/resources/pod/checker.go#L313. 

```go
func (c *controller) PatrollerDo() {
	// step1: get pList
	pList, err := c.podLister.List(util.GetSuperClusterListerLabelsSelector())

	for _, cluster := range clusterNames {
		// step2: get vList
		vList := &corev1.PodList{}
		if err := c.MultiClusterController.List(cluster, vList); err != nil {
			// ...
		}
	}
}
```
Here, we get the pList  and vList  cache in two steps with different time. As the number of clusters become more, there will be a greater difference.  Some pPods will not be visible when Pods created during Checker run.  This will trigger the pod force delete logic.

```go
	// pPod not found and vPod still exists, the pPod may be deleted manually or by controller pod eviction.
	// If the vPod has not been bound yet, we can create pPod again.
	// If the vPod has been bound, we'd better delete the vPod since the new pPod may have a different nodename.
	if isPodScheduled(vPod) {
		c.forceDeleteVPod(vObj.GetOwnerCluster(), vPod, false)
		return
	}
```

Meanwhile, Pod DWS syncer will trigger deletion of pPod.

When finding that a vPod has been bound without pPod, we should double check whether the pPod exists before the pod force deletion .

/kind bug

wondywang:
Because getting p/v objects in checker are not atomic operations. Here, we get the pList and vList cache in two steps with different time. As the number of clusters become more, there will be a greater difference. Some pPods will not be visible when Pods created during Checker run. This will trigger the pod force delete logic.

So, when finding that a vPod has been bound without pPod, we need double check whether the pPod exists before the pod force deletion.

**Which issue(s) this PR fixes** :
Fixes #308

### DIFF
--- a/virtualcluster/pkg/syncer/resources/pod/checker.go
+++ b/virtualcluster/pkg/syncer/resources/pod/checker.go
@@ -311,6 +311,12 @@ func (c *controller) differAddFunc(vObj differ.ClusterObject) {
 	// If the vPod has not been bound yet, we can create pPod again.
 	// If the vPod has been bound, we'd better delete the vPod since the new pPod may have a different nodename.
 	if isPodScheduled(vPod) {
+		// double check pPod exist or not.
+		targetNamespace := conversion.ToSuperClusterNamespace(vObj.OwnerCluster, vPod.Namespace)
+		if _, err := c.podLister.Pods(targetNamespace).Get(vPod.Name); !apierrors.IsNotFound(err) {
+			klog.Warningf("pPod %s found by double check, should not delete vPod anymore", vObj.Key)
+			return
+		}
 		c.forceDeleteVPod(vObj.GetOwnerCluster(), vPod, false)
 		metrics.CheckerRemedyStats.WithLabelValues("DeletedTenantPodsDueToSuperEviction").Inc()
 		return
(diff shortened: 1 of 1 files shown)


virtualcluster/pkg/syncer/resources/pod/checker.go
```diff
@@ -311,6 +311,12 @@ func (c *controller) differAddFunc(vObj differ.ClusterObject) {
 	// If the vPod has not been bound yet, we can create pPod again.
 	// If the vPod has been bound, we'd better delete the vPod since the new pPod may have a different nodename.
 	if isPodScheduled(vPod) {
+		// double check pPod exist or not.
+		targetNamespace := conversion.ToSuperClusterNamespace(vObj.OwnerCluster, vPod.Namespace)
+		if _, err := c.podLister.Pods(targetNamespace).Get(vPod.Name); !apierrors.IsNotFound(err) {
```
Fei-Guo, line 316: Can you do a nil check for the err first before calling apierrors.IsNotFound?

wondywang: OK, it's done.

Fei-Guo (approved):

PR #313, opened 2022-09-27
Fix dws reconcile should return true by default

wondywang:
dws reconcile should return true by default. Otherwise, the workflow will jitter, with brief pauses.

Fixes #

### DIFF
--- a/virtualcluster/pkg/util/mccontroller/mccontroller.go
+++ b/virtualcluster/pkg/util/mccontroller/mccontroller.go
@@ -474,7 +474,7 @@ func (c *MultiClusterController) processNextWorkItem() bool {
 	metrics.RecordDWSOperationStatus(c.objectKind, req.ClusterName, utilconstants.StatusCodeError)
 	c.Queue.AddRateLimited(req)
 	klog.Errorf("%s dws request reconcile failed: %v", req, err)
-	return false
+	return true
 }
 
 func (c *MultiClusterController) FilterObjectFromSchedulingResult(req reconciler.Request) bool {


wondywang:
PTAL @christopherhein

christopherhein:
@wondywang Wouldn't jittering be expected from this so that we don't continue to retry failed queue operations? Maybe I'm mistaken by what this part was expecting to do.

cc @Fei-Guo you might have better context.

wondywang:
> 
thanks @christopherhein , of course, failed queue operations would be re-queued for processing, `MultiClusterController` work like this:

```go
func (c *MultiClusterController) Start(stop <-chan struct{}) error {
	for i := 0; i < c.MaxConcurrentReconciles; i++ {
		// if worker return, it will run after JitterPeriod.
		go wait.Until(c.worker, c.JitterPeriod, stop)
	}
	return nil
}

func (c *MultiClusterController) worker() {
	// here we want processNextWorkItem always return true, unless the queue is shutdown.
	for c.processNextWorkItem() {
	}
}

func (c *MultiClusterController) processNextWorkItem() bool {
	if shutdown {
		// Stop working
		return false
	}

	// do some process, if failed, re-enqueue.

	return true
}
```
so, if we return false at  the end of func `processNextWorkItem`, controller will stop a moment. This will be a long waiting of time if we encounter a large number of failures.

Fei-Guo:
> ```go
> processNextWorkItem
> ```

If you check the source code of Kubernetes controller such as https://github.com/kubernetes/kubernetes/blob/e450d35bb303042830a8ebfaad2010908df61208/pkg/controller/daemon/daemon_controller.go#L329, processNextWorkItem returns true in cases of failures. I think it works because we add rate limited for the failed keys hence, we are not doing busy retries for failures.

christopherhein:
Gotcha, so we should go forward and approve this then, since we're not doing the right flow then to keep in-sync with expected controllers. Thanks @Fei-Guo

PR #314, opened 2022-09-27
✨ Convert PodMutators to Plugins

m-messiah:
**What this PR does / why we need it**: The PR converts PodMutators mechanism to Plugins, allowing any additional vc-syncer implementations to easily add their own pod mutator methods. The PR defines a simple mutatorplugin.Interface to implement and converts two existing mutators (ServiceLink and ServiceAccountToken) to plugins.
The default pod mutator uses too many controller methods and caches, so it is not worth time to convert it to be a plugin, especially, while we want it to be executed first to be the actual "default". I could try to migrate it as a follow-up if you think it is needed.

I also extended PodMutateCtx to have vPod too, to simplify comparisons and mutations in the future plugins.

Relate: https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/311#issuecomment-1259073854


Fei-Guo:
Looks reasonable to me. 

@wondywang ,  can you take a look as well?

wondywang:
> 

/lgtm
thanks @m-messiah

christopherhein (approved):
/lgtm
/approve

Thanks @m-messiah !

Issue #63, opened 2021-05-19
Unable to create a VirtualCluster on k8 v1.20.2


mazocode:
## Problem

Virtual cluster does not deploy with k8 v1.20.2. Output from vc-manager:

> {"level":"info","ts":1621457612.1333222,"logger":"clusterversion-controller","msg":"reconciling ClusterVersion..."}
> {"level":"info","ts":1621457612.1334903,"logger":"clusterversion-controller","msg":"new ClusterVersion event","ClusterVersionName":"cv-sample-np"}
> {"level":"info","ts":1621457635.4177175,"logger":"virtualcluster-webhook","msg":"validate create","vc-name":"vc-sample-1"}
> {"level":"info","ts":1621457635.4421399,"logger":"virtualcluster-controller","msg":"reconciling VirtualCluster..."}
> {"level":"info","ts":1621457635.4824774,"logger":"virtualcluster-webhook","msg":"validate update","vc-name":"vc-sample-1"}
> {"level":"info","ts":1621457635.511791,"logger":"virtualcluster-controller","msg":"a finalizer has been registered for the VirtualCluster CRD","finalizer":"virtualcluster.finalizer.native"}
> {"level":"info","ts":1621457635.5118568,"logger":"virtualcluster-controller","msg":"will create a VirtualCluster","vc":"vc-sample-1"}
> {"level":"info","ts":1621457635.53576,"logger":"virtualcluster-webhook","msg":"validate update","vc-name":"vc-sample-1"}
> {"level":"info","ts":1621457635.556264,"logger":"virtualcluster-controller","msg":"reconciling VirtualCluster..."}
> {"level":"info","ts":1621457635.5563915,"logger":"virtualcluster-controller","msg":"VirtualCluster is pending","vc":"vc-sample-1"}
> {"level":"info","ts":1621457638.3632772,"logger":"virtualcluster-controller","msg":"creating secret","name":"root-ca","namespace":"default-a4a766-vc-sample-1"}
> {"level":"info","ts":1621457638.400915,"logger":"virtualcluster-controller","msg":"creating secret","name":"apiserver-ca","namespace":"default-a4a766-vc-sample-1"}
> {"level":"info","ts":1621457638.4276915,"logger":"virtualcluster-controller","msg":"creating secret","name":"etcd-ca","namespace":"default-a4a766-vc-sample-1"}
[...]

PR #316, opened 2022-10-05
🐛Fix VirtualCluster CRD and reconciler for work in 1.19+1.20 environment

m-messiah:
When the super cluster has two different versions of API Server (specifically 1.19 and 1.20) it results in errors of vc-manager, as 1.20 added non-nullable nulls https://github.com/kubernetes/kubernetes/pull/95423, while 1.19 can't work with them.

```
{"level":"error","ts":1664969234.2211943,"logger":"controller-runtime.manager.controller.virtualcluster","msg":"Reconciler error","reconciler group":"tenancy.x-k8s.io","reconciler kind":"VirtualCluster","name":"test","namespace":"default","error":"VirtualCluster.tenancy.x-k8s.io \"test\" is invalid: [status.reason: Invalid value: \"null\": status.reason in body must be of type string: \"null\", status.message: Invalid value: \"null\": status.message in body must be of type string: \"null\", status.phase: Invalid value: \"null\": status.phase in body must be of type string: \"null\"]"}

{"level":"error","ts":1664975149.3489153,"logger":"controller-runtime.manager.controller.virtualcluster","msg":"Reconciler error","reconciler group":"tenancy.x-k8s.io","reconciler kind":"VirtualCluster","name":"test","namespace":"default","error":"VirtualCluster.tenancy.x-k8s.io \"test\" is invalid: status.conditions.status: Invalid value: \"null\": status.conditions.status in body must be of type string: \"null\""}
```

So the first solution is to set LastTransitionTime to be a pointer (as metav1.Time is not nullable https://github.com/kubernetes/kubernetes/issues/86811
And the second solution is to set status.condition.Status to non-empty string (as it expected initially). I set it always True, as we don't have false conditions, but we could improve it later.

With these two changes vc-manager became working again.

The PR is related to https://github.com/kubernetes-sigs/cluster-api-provider-nested/issues/63, but I am not sure it could fix all the things there


Fei-Guo:
I am a little confused about the problematic version combination of vc, vc-manager, super. Can you provide the details of the following?

vc apiserver version: ?
vc-manager client-go version: ?
super apiserver version: ?

m-messiah:
vc apiserver version: does not matter. Happens with 1.19 and 1.20 as well
vc-manager client-go version: latest commit https://github.com/kubernetes-sigs/cluster-api-provider-nested/commit/e55cc6f87f61edc1fa028c4ba11a69c5e5b4330a ,  v0.21.9 as I see in go.mod
super apiserver version: v1.19.13 + v1.20.15 on several nodes


So the main problem here is we have two apiserver versions in the super cluster during migration upgrade, that leads to incorrectly set CRs (by api 1.20), that are needed to be handled by 1.19.13 apiserver.

Fei-Guo:
@christopherhein Yes, this is crd field type change without bumping the version. I am not sure if existing CRs need to be converted to make controller or cr list/get work.

m-messiah:
We found that the problem was deeper than it looked, especially after additional checks that this PR changes did not help. 
The root cause of the issue is described here https://github.com/kubernetes/kubernetes/issues/99804 and fixed there https://github.com/kubernetes/kube-openapi/pull/230 and it happened for our cluster where we had couple Kubernetes API Servers built with Golang 1.16, that failed the validation. 

The fast and only solution is to rebuilt the server with Go 1.15, as only Kube APIServer 1.21.0+ has the bugfix for 1.16 compatibility included

PR #317, opened 2022-10-06
✨Bypass vc-syncing if tenant has own scheduler and dedicated node.

weiling61:
For certain tenants, they may be assigned with dedicated nodes and can schedule workload (pod) to these nodes with tenant scheduler. We would like to introduce a new label: "tenancy.x-k8s.io/tenant-scheduler" to let vc-syncer to be aware of the situation and skip the syncing.

Fixes # N/A

### DIFF
--- a/virtualcluster/pkg/syncer/constants/constants.go
+++ b/virtualcluster/pkg/syncer/constants/constants.go
@@ -64,7 +64,7 @@ const (
 	// LabelSecretUID is the service account token secret UID in tenant namespace.
 	LabelSecretUID = "tenancy.x-k8s.io/secret.UID" // #nosec G101 -- This is a label key
 
-	// LabelTenantSchedulerName is used to by resources that do not need to be synced, but to be scheduled to a tenant controled node
+	// LabelTenantSchedulerName is used to by resources that do not need to be synced, but to be scheduled to a tenant controlled node
 	LabelTenantSchedulerName = "tenancy.x-k8s.io/tenant-scheduler"
 
 	// UwsControllerWorkerHigh is the quantity of the worker routine for a resource that generates high number of uws requests.
--- a/virtualcluster/pkg/syncer/constants/constants.go
+++ b/virtualcluster/pkg/syncer/constants/constants.go
@@ -64,6 +64,9 @@ const (
 	// LabelSecretUID is the service account token secret UID in tenant namespace.
 	LabelSecretUID = "tenancy.x-k8s.io/secret.UID" // #nosec G101 -- This is a label key
 
+	// LabelTenantSchedulerName is used to by resources that do not need to be synced, but to be scheduled to a tenant controled node
+	LabelTenantSchedulerName = "tenancy.x-k8s.io/tenant-scheduler"
+
 	// UwsControllerWorkerHigh is the quantity of the worker routine for a resource that generates high number of uws requests.
 	UwsControllerWorkerHigh = 10
 	// UwsControllerWorkerLow is the quantity of the worker routine for a resource that generates low number of uws requests.
--- a/virtualcluster/pkg/syncer/constants/constants.go
+++ b/virtualcluster/pkg/syncer/constants/constants.go
@@ -64,7 +64,7 @@ const (
 	// LabelSecretUID is the service account token secret UID in tenant namespace.
 	LabelSecretUID = "tenancy.x-k8s.io/secret.UID" // #nosec G101 -- This is a label key
 
-	// LabelTenantIgnoreSync is used by resources that do not need to be synced, but to be scheduled to a tenant controlled node
+	// LabelTenantIgnoreSync is used by resources that do not need to be synced.
 	LabelTenantIgnoreSync = "tenancy.x-k8s.io/ignore-sync"
 
 	// UwsControllerWorkerHigh is the quantity of the worker routine for a resource that generates high number of uws requests.
--- a/virtualcluster/pkg/syncer/constants/constants.go
+++ b/virtualcluster/pkg/syncer/constants/constants.go
@@ -64,8 +64,8 @@ const (
 	// LabelSecretUID is the service account token secret UID in tenant namespace.
 	LabelSecretUID = "tenancy.x-k8s.io/secret.UID" // #nosec G101 -- This is a label key
 
-	// LabelTenantSchedulerName is used to by resources that do not need to be synced, but to be scheduled to a tenant controlled node
-	LabelTenantSchedulerName = "tenancy.x-k8s.io/tenant-scheduler"
+	// LabelTenantIgnoreSync is used by resources that do not need to be synced, but to be scheduled to a tenant controlled node
+	LabelTenantIgnoreSync = "tenancy.x-k8s.io/ignore-sync"
 
 	// UwsControllerWorkerHigh is the quantity of the worker routine for a resource that generates high number of uws requests.
 	UwsControllerWorkerHigh = 10
--- a/virtualcluster/pkg/syncer/resources/pod/checker.go
+++ b/virtualcluster/pkg/syncer/resources/pod/checker.go
@@ -160,7 +160,7 @@ func (c *controller) PatrollerDo() {
 	vSet := differ.NewDiffSet()
 
 	sel := labels.NewSelector()
-	if featuregate.DefaultFeatureGate.Enabled(featuregate.ResourceNoSync) {
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.TenantAllowResourceNoSync) {
 		r, err := labels.NewRequirement(constants.LabelTenantIgnoreSync, selection.NotEquals, []string{"true"})
 		if err == nil {
 			sel = sel.Add(*r)
--- a/virtualcluster/pkg/syncer/resources/pod/checker.go
+++ b/virtualcluster/pkg/syncer/resources/pod/checker.go
@@ -170,6 +170,12 @@ func (c *controller) PatrollerDo() {
 					continue
 				}
 			}
+			// if LabelTenantSchedulerName is specified, bypass syncing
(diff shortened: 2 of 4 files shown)


virtualcluster/pkg/syncer/resources/pod/checker.go
```diff
@@ -170,6 +170,12 @@ func (c *controller) PatrollerDo() {
 					continue
 				}
 			}
+			// if LabelTenantSchedulerName is specified, bypass syncing
+			schedulerlabel, ok := vList.Items[i].GetLabels()[constants.LabelTenantSchedulerName]
+			if ok && schedulerlabel != "" {
+				klog.V(5).Infof("skip syncing pod with tenant scheduler label %v", schedulerlabel)
+				continue
+			}
```
christopherhein, line 178: We have a func in the checker already which will filter out pods with a label rather than iterating since this is expensive - https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/e7f1762df3752d2e1206ff06b23aed6ff3c8bd40/virtualcluster/pkg/syncer/resources/pod/checker.go#L146-L150 maybe we can change this to also check for a tenant scheduler label if there is a feature-gate turning on this.

virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -64,6 +64,12 @@ func (c *controller) Reconcile(request reconciler.Request) (res reconciler.Resul
 	if err := c.MultiClusterController.Get(request.ClusterName, request.Namespace, request.Name, vPod); err != nil && !apierrors.IsNotFound(err) {
 		return reconciler.Result{Requeue: true}, err
 	}
+	// if LabelTenantSchedulerName is specified, bypass syncing
+	schedulerlabel, ok := vPod.GetLabels()[constants.LabelTenantSchedulerName]
+	if ok && schedulerlabel != "" {
+		klog.V(5).Infof("skip syncing pod with tenant scheduler label")
+		return reconciler.Result{}, nil
+	}
```
christopherhein, line 72: If you add this logic - https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/pkg/syncer/resources/pod/controller.go#L317-L319 it only even enqueue those pods so this check won't be necessary to fire, instead the informer will ignore it, it also needs to be behind a feature gate IMO.

christopherhein (changes requested):
Couple comments to make this more platform level opt-in and move the checks to places that would be more efficient.

virtualcluster/pkg/syncer/resources/pod/checker.go
weiling61: can not use pList, since pod is not going to be synced to super cluster, so only check for vList

christopherhein: Sorry ignore this, as you pointed out this is for the vPod not the pPod which that line implements.

christopherhein: I see, yea my mistake, can we do that filtering here - https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/e7f1762df3752d2e1206ff06b23aed6ff3c8bd40/virtualcluster/pkg/syncer/resources/pod/checker.go#L160-L164 then.

christopherhein: Which is what sets up the `differ`, so rather than loading all vPods it will filter from the informer those pods with this label (only when the feature gate is enabled)

weiling61: k

christopherhein:
I am good with this since it's opt-in @Fei-Guo @wondywang would either of you be able to check this out as well?

christopherhein (approved):

Fei-Guo:
Can you explain more about the background?  Are you directly registering nodes to tenant control plane? 

In the virtual cluster model, even if tenant wants to use dedicated nodes, the nodes should be registered to super cluster and exposed to tenant control plane using vNodes.

If you directly register node to tenant control plane, I'd rather you just introduce a generic feature to mark vPod not to be synchronized.

virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -64,6 +64,14 @@ func (c *controller) Reconcile(request reconciler.Request) (res reconciler.Resul
 	if err := c.MultiClusterController.Get(request.ClusterName, request.Namespace, request.Name, vPod); err != nil && !apierrors.IsNotFound(err) {
 		return reconciler.Result{Requeue: true}, err
 	}
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.TenantScheduler) {
```
m-messiah, line 67: let's move `pPod, err := c.podLister.Pods(targetNamespace).Get(request.Name)` after this check to skip unneeded superCluster pod lookup?

virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -64,6 +64,14 @@ func (c *controller) Reconcile(request reconciler.Request) (res reconciler.Resul
...
+		// if LabelTenantSchedulerName is specified, bypass syncing
+		schedulerlabel, ok := vPod.GetLabels()[constants.LabelTenantSchedulerName]
+		if ok && schedulerlabel != "" {
```
m-messiah, line 70: you can omit ok, with just `if vPod.GetLabels()[constants.LabelTenantSchedulerName] != ""` as it is "" if not found

virtualcluster/pkg/syncer/resources/pod/dws.go
weiling61: ok

virtualcluster/pkg/syncer/resources/pod/dws.go
weiling61: Yes true, but I prefer to play the safe ...

weiling61:
@Fei-Guo "Can you explain more about the background? Are you directly registering nodes to tenant control plane?"
[answer] Yes I directly registered nodes for some Virtual Clusters and added a scheduler for each virtual cluster.

"In the virtual cluster model, even if tenant wants to use dedicated nodes, the nodes should be registered to super cluster and exposed to tenant control plane using vNodes."
[answer] This is still an experimental feature. Directly registering node to virtual cluster brings other advantages. In near future, we may propose to block other resources from syncing to super cluster.

"If you directly register node to tenant control plane, I'd rather you just introduce a generic feature to mark vPod not to be synchronized."
[answer] It is not just about vPod. So for this radar, I add an experimental feature gate, to just make things move forward. We may propose a more generic solution once I completed prototype.

virtualcluster/pkg/syncer/constants/constants.go
```diff
@@ -64,6 +64,9 @@ const (
 	// LabelSecretUID is the service account token secret UID in tenant namespace.
 	LabelSecretUID = "tenancy.x-k8s.io/secret.UID" // #nosec G101 -- This is a label key
 
+	// LabelTenantSchedulerName is used to by resources that do not need to be synced, but to be scheduled to a tenant controlled node
+	LabelTenantSchedulerName = "tenancy.x-k8s.io/tenant-scheduler"
```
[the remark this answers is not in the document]
weiling61: I need to use this label to prevent other resources being synced to super cluster in next step. This is why I did not pick label name specific to "pod"....

Fei-Guo:
@weiling61 

You can choose label name like "tenancy.x-k8s.io/ignore-sync=true" and just make a feature to prevent synchronizing any objects.  The feature name "TenantScheduler" does not make much sense for other non-Pod objects anyway.

weiling61:
@Fei-Guo Change the new label suggested in your comments

virtualcluster/pkg/syncer/resources/pod/dws.go
```diff
@@ -55,15 +55,22 @@ func (c *controller) Reconcile(request reconciler.Request) (res reconciler.Resul
...
-		return reconciler.Result{Requeue: true}, err
-	}
-
 	vPod := &corev1.Pod{}
 	if err := c.MultiClusterController.Get(request.ClusterName, request.Namespace, request.Name, vPod); err != nil && !apierrors.IsNotFound(err) {
 		return reconciler.Result{Requeue: true}, err
 	}
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.TenantScheduler) {
+		// if constants.LabelTenantIgnoreSync is true, bypass syncing
+		schedulerlabel, ok := vPod.GetLabels()[constants.LabelTenantIgnoreSync]
```
Fei-Guo, line 64: can you rename the variable name "schedulerlabel"? we should not imply scheduling anywhere.

weiling61: Sure

virtualcluster/pkg/syncer/util/featuregate/gate.go
```diff
@@ -68,6 +68,10 @@ const (
 	// ClusterVersionPartialUpgrade is an experimental feature that allows the cluster provisioner
 	// to apply ClusterVersion updates if VirtualCluster object is requested it
 	ClusterVersionPartialUpgrade = "ClusterVersionPartialUpgrade"
+
+	// ResourceNoSync is an experimental feature that gives tenant the capability
+	// of not syncing certain resources to super cluster.
+	ResourceNoSync = "ResourceNoSync"
```
Fei-Guo, line 74: Sorry for being paranoid, I wish the feature name exactly expresses its intention. Hence I suggest naming it to
"TenantAllowResourceNoSync"

weiling61: This name is better. Thanks

PR #319, opened 2022-10-10
🐛Fix Pod Mutator ordering

### LINKED ISSUE #318: 🐛 Pod Mutator has order requirements
Because the `PodServiceLinkMutatorPlugin` PodMutator is executed later than `PodMutateDefault`, the enableServiceLinks switch does not take effect, and the Services in the namespace where the pPod is located is injected into the Pod as an environment variable.

```go
func (c *controller) reconcilePodCreate(clusterName, targetNamespace, requestUID string, vPod *corev1.Pod) error {
	// ...

	var ms = append([]conversion.PodMutator{
		conversion.PodMutateDefault(vPod, pSecretMap, services, nameServer, c.Config.DNSOptions),
	}, c.podMutators...)

	err = conversion.VC(c.MultiClusterController, clusterName).Pod(pPod, vPod).Mutate(ms...)
	// ...
}

func PodMutateDefault(vPod *v1.Pod, saSecretMap map[string]string, services []*v1.Service, nameServer string, dnsOption []v1.PodDNSConfigOption) PodMutator {
	return func(p *PodMutateCtx) error {

		// here, pPod.Spec.EnableServiceLinks determines whether Services need to be injected as environment variables.
		apiServerClusterIP, serviceEnv := getServiceEnvVarMap(p.PPod.Namespace, p.ClusterName, p.PPod.Spec.EnableServiceLinks, services)

		// ...

		return nil
	}
}
```
Pod Mutator has order requirements. Pod.Spec.EnableServiceLinks needs to be set before PodMutateDefault. 

`PodServiceLinkMutatorPlugin` PodMutator run before `PodMutateDefault`, or PodMutator executes in order.

/kind bug

m-messiah:
**What this PR does / why we need it**: PodServiceLinkMutatorPlugin should be executed before the Default PodMutator (as well as PodMountServiceAccountTokenMutator). The first assumption in https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/314 was incorrect and we need to invoke Default after all mutations, rather than before. In the ideal way, we should have Default mutator in the same list of ordering, being able to execute some plugins before and some after the default, but Default Mutator uses too much controller fields (as serviceLister and secretLister), so it is hard to migrate it somewhere else without changing PodMutateCtx having all controller things. We should consider doing it later when it will be really needed.
The PR also adds the missing test for serviceLinks

Fixes #318

### DIFF
--- a/virtualcluster/pkg/syncer/resources/pod/dws.go
+++ b/virtualcluster/pkg/syncer/resources/pod/dws.go
@@ -197,12 +197,9 @@ func (c *controller) reconcilePodCreate(clusterName, targetNamespace, requestUID
 		return fmt.Errorf("failed to find nameserver: %v", err)
 	}
 
-	// We need to apply default mutator first
-	var ms = append([]conversion.PodMutator{
-		// TODO: Convert PodMutateDefault to a plugin
-		// It is not an easy task as it uses a lot of controller methods now, but could be nice to be generalised.
-		conversion.PodMutateDefault(vPod, pSecretMap, services, nameServer, c.Config.DNSOptions),
-	}, c.podMutators...)
+	// TODO: Convert PodMutateDefault to a plugin
+	// It is not an easy task as it uses a lot of controller methods now, but could be nice to be generalised.
+	var ms = append(c.podMutators, conversion.PodMutateDefault(vPod, pSecretMap, services, nameServer, c.Config.DNSOptions))
 
 	err = conversion.VC(c.MultiClusterController, clusterName).Pod(pPod, vPod).Mutate(ms...)
 	if err != nil {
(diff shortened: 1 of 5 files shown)


wondywang:
LGTM. But, I am not quite sure  if the change of ResourceRegister List is suitable.  

@Fei-Guo can you take a look as well?

christopherhein:
/lgtm

But I will let Fei approve/comment. Needing ordering is a lot like how the admission registration works for api servers, based on naming to denote ordering.

Fei-Guo:
If plugin order matters, can we somehow add a test to check order?

m-messiah:
> If plugin order matters, can we somehow add a test to check the order?

simple test added

PR #321, opened 2022-10-15
🐛 Handle deprecated preserveUnknownField in CRD spec

kaushik229:
In Kubernetes 1.20 the spec.preserveUnknownFields is not allowed to be set to true given Kubernetes has already deprecated this any cluster >=1.20 will not be able sync these resources, instead if a CRD has the field still set this disables that.

### DIFF
--- a/virtualcluster/pkg/syncer/conversion/helper.go
+++ b/virtualcluster/pkg/syncer/conversion/helper.go
@@ -306,6 +306,14 @@ func BuildVirtualPriorityClass(cluster string, pPriorityClass *v1scheduling.Prio
 func BuildVirtualCRD(cluster string, pCRD *apiextensionsv1.CustomResourceDefinition) *apiextensionsv1.CustomResourceDefinition {
 	vCRD := pCRD.DeepCopy()
 	ResetMetadata(vCRD)
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.DisableCRDPreserveUnknownFields) {
+		// In Kubernetes 1.20 the spec.preserveUnknownFields is not allowed to be set to true
+		// given Kubernetes has already deprecated this any cluster >=1.20 will not be able
+		// sync these resources, instead if a CRD has the field still set this disables that.
+		vCRD.Spec.PreserveUnknownFields = false
+
+	}
+
 	return vCRD
 }
 
--- a/virtualcluster/pkg/syncer/util/featuregate/gate.go
+++ b/virtualcluster/pkg/syncer/util/featuregate/gate.go
@@ -72,18 +72,23 @@ const (
 	// TenantAllowResourceNoSync is an experimental feature that gives tenant the capability
 	// of not syncing certain resources to super cluster.
 	TenantAllowResourceNoSync = "TenantAllowResourceNoSync"
+
+	// DisableCRDPreserveUnknownFields helps control syncing deprecated(k8s <= 1.20) field on CRD.
+	// Enabling this will set spec.preserveUnknownField to false regardless the value in source CRD spec.
+	DisableCRDPreserveUnknownFields = "DisableCRDPreserveUnknownFields"
 )
 
 var defaultFeatures = FeatureList{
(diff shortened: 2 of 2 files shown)


christopherhein (commented):
/lgtm

@Fei-Guo, given that field is deprecated, do you see issues in force setting this on the CRD syncer.

christopherhein:
/retitle 🐛 Handle deprecated preserveUnknownField in CRD spec

Fei-Guo:
I'd like to understand the case better.  The problem will happen when super cluster runs k8s < 1.20 and vc runs k8s > 1.20.  Should we support this? I believe we will have more compatibility issues if vc runs higher version than super.

christopherhein:
The use case is a super cluster running 1.20 with a CRD that was created under the v1beta1 API where `preserveUnknownField: true`, and having it labeled to sync into another tenant control plane running v1.20, the object comes across with the deprecated field sent to true and will fail creation because of this.

The "easy" answer is set the super cluster version to `false`, which I 100% agree with, but where this causes issues is if folks use unknown fields their resources fail creation.

Fei-Guo:
> The use case is a super cluster running 1.20 with a CRD that was created under the v1beta1 API where `preserveUnknownField: true`, and having it labeled to sync into another tenant control plane running v1.20, the object comes across with the deprecated field sent to true and will fail creation because of this.
> 
> The "easy" answer is set the super cluster version to `false`, which I 100% agree with, but where this causes issues is if folks use unknown fields their resources fail creation.

Sure, but on the other hand, this change may silently create discrepancy between vc cr and super cr - the persisted data can be different.  Shouldn't we suggest user to resolve the super crd first? If the unknown fields are necessary, just adding the validation schemas should be sufficient, right?

kaushik229:
@Fei-Guo I'm new to this and was reading up a little on this.
It seems this is little tricky but few recommendation I found asks to set it to false.
https://github.com/kubernetes/kubernetes/issues/95702#issuecomment-713548542

Fei-Guo:
> Definitely true, but also in that case the CRDs won't ever be synced at all cause they will fail the CRD creation validation in the tenant cluster.

My suggestion is to set the field to false and correct all validation schema if unknown fields are necessary in super CRD before syncing it to tenant. We force users to do this which I believe is a good experience from management perspective. Do you think this is impractical?  Maybe there are some other blocking factors that I am not aware of.

christopherhein:
For us its a management issue, tbh, we can't update our CRD without breaking existing clients and we can't get these CRDs to sync so folks can use "valid" CRDs. Our alternative is to disable this syncer and manually sync the CRD with the `spec.preserveUnknownFields: false` in the tenant clusters.

virtualcluster/pkg/syncer/conversion/helper.go
```diff
@@ -306,6 +306,15 @@ func BuildVirtualPriorityClass(cluster string, pPriorityClass *v1scheduling.Prio
 func BuildVirtualCRD(cluster string, pCRD *apiextensionsv1.CustomResourceDefinition) *apiextensionsv1.CustomResourceDefinition {
 	vCRD := pCRD.DeepCopy()
 	ResetMetadata(vCRD)
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.SyncCRDPreserveUnknownField) {
+		// In Kubernetes 1.20 the spec.preserveUnknownFields is not allowed to be set to true
+		// given Kubernetes has already deprecated this any cluster >=1.20 will not be able
+		// sync these resources, instead if a CRD has the field still set this disables that.
+		if vCRD.Spec.PreserveUnknownFields == true {
+			vCRD.Spec.PreserveUnknownFields = false
+		}
```
christopherhein, line 315: We can get rid of the additional `if` just always set to `false`

kaushik229: done

christopherhein (approved):
/lgtm
/approve

Thank you!

PR #323, opened 2022-10-26
✨ Add PersistentVolumeClaim status UWS

### LINKED ISSUE #322: ✨PersistentVolumeClaim support UWS status update
**User Story**

As a user, I want to update the spec.resources.requests.storage of PVC, but I found that after updating requests.storage, status.capacity.storage is not updated synchronously, resulting in an exception when kubectl get pvc.

**Detailed Description**

image

Should we consider updating the pPVC status synchronously to the pPVC?

/kind feature

wondywang:

Add PersistentVolumeClaim status upward sync. When we resize the storage of vPVC in VC, the CSI driver on the super cluster would update the pPVC.status.capacity after capacity expansion. Then we need to update vPVC status synchronously. 

This time, we only consider the capacity update, and other information (such as accessModes and phase) will not be synchronized.

Relate: #322

/kind feature

### DIFF
--- a/virtualcluster/pkg/syncer/resources/persistentvolumeclaim/uws.go
+++ b/virtualcluster/pkg/syncer/resources/persistentvolumeclaim/uws.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/virtualcluster/pkg/syncer/resources/persistentvolumeclaim/uws.go
+++ b/virtualcluster/pkg/syncer/resources/persistentvolumeclaim/uws.go
@@ -19,6 +19,7 @@ package persistentvolumeclaim
 import (
 	"context"
 	"fmt"
+
 	pkgerr "github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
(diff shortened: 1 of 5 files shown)


wondywang:
PTAL @christopherhein

Fei-Guo:
1) Please change the copyright year of all touched/new files to 2022.
2) You need to modify the patroller to check the status consistency for PVC (can be in another PR if you prefer).

virtualcluster/pkg/syncer/resources/persistentvolumeclaim/uws.go
```diff
@@ -0,0 +1,83 @@
...
+
+	clusterName, vNamespace := conversion.GetVirtualOwner(pPVC)
+	if clusterName == "" {
+		// Bound PVC does not belong to any tenant.
+		return nil
+	}
+
+	tenantClient, err := c.MultiClusterController.GetClusterClient(clusterName)
+	if err != nil {
+		return pkgerr.Wrapf(err, "failed to create client from cluster %s config", clusterName)
```
m-messiah, line 65: there is no need to use github.com/pkg/errors since golang 1.13 (for example https://stackoverflow.com/a/61955455/1479097)

```suggestion
		return fmt.Errorf("failed to create client from cluster %s config: %w", clusterName, err)
```

m-messiah (changes requested):
please, remove github.com/pkg/errors usage

virtualcluster/pkg/syncer/resources/persistentvolumeclaim/uws.go
wondywang: thx, have changed.

m-messiah (approved):
/lgtm
/approve

wondywang:
> 1. Please change the copyright year of all touched/new files to 2022.
> 2. You need to modify the patroller to check the status consistency for PVC (can be in another PR if you prefer).

thx, have done.

PR #326, opened 2022-11-08
🐛 v1.21 Support for RootCACertConfigMap feature

### LINKED ISSUE #293: Resource already exists and the UID is different should not requeue

christopherhein:
This PR adds feature gated support for VirtualCluster to work with any Kubernetes version +1.21 and when the `RootCACertConfigMap` feature is enabled in 1.20 clusters. To implement this rather than fighting the super cluster for the `kube-root-ca.crt` we will create a prefixed ConfigMap in each namespace `tenant-kube-root-ca.crt` and then when pods are synced we have an additional mutator that will rewrite the ConfigMap to the proper resource. This allows both super and tenant to work and requires no changes from a tenants workloads.

Fixes #293

### DIFF
--- a/virtualcluster/pkg/syncer/conversion/equality.go
+++ b/virtualcluster/pkg/syncer/conversion/equality.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/apis/tenancy/v1alpha1"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/apis/config"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/constants"
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/featuregate"
 )
 
 type vcEquality struct {
@@ -447,6 +448,10 @@ func (e vcEquality) checkInt64Equality(pObj, vObj *int64) (*int64, bool) {
 // CheckConfigMapEquality checks whether super control plane ConfigMap and virtual ConfigMap
 // are logically equal. The source of truth is virtual object.
 func (e vcEquality) CheckConfigMapEquality(pObj, vObj *v1.ConfigMap) *v1.ConfigMap {
+	// Before processing Equality on ConfigMaps we need to reset the CM if pObj is used for
+	// RootCACertConfigMaps.
+	_, pObj.Name = GetConfigMapName(vObj.Name)
+
 	var updated *v1.ConfigMap
 	updatedMeta := e.CheckDWObjectMetaEquality(&pObj.ObjectMeta, &vObj.ObjectMeta)
 	if updatedMeta != nil {
@@ -470,6 +475,13 @@ func (e vcEquality) CheckConfigMapEquality(pObj, vObj *v1.ConfigMap) *v1.ConfigM
 			updated = pObj.DeepCopy()
 		}
 		updated.BinaryData = updateBinaryData
+	}
+
+	// After we have processed the equality checks if updated is not nil reupdate to include
+	// super's ConfigMap Name
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.RootCACertConfigMapSupport) &&
+		vObj.Name == constants.RootCACertConfigMapName && updated != nil {
+		updated.Name = constants.TenantRootCACertConfigMapName
 	}
 
 	return updated
--- a/virtualcluster/pkg/syncer/resources/configmap/checker.go
+++ b/virtualcluster/pkg/syncer/resources/configmap/checker.go
@@ -33,6 +33,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/metrics"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/patrol/differ"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util"
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/featuregate"
 )
 
 var numMissMatchedConfigMaps uint64
@@ -63,6 +64,13 @@ func (c *controller) PatrollerDo() {
 	}
 	pSet := differ.NewDiffSet()
 	for _, pCM := range pConfigMaps {
+		// Ingore both RootCACertConfigMapName and TenantRootCACertConfigMapName from the super.
+		// TenantRootCACertConfigMapName is created from the vRootCACertConfigMap and
+		// RootCACertConfigMapName should be left alone.
+		if featuregate.DefaultFeatureGate.Enabled(featuregate.RootCACertConfigMapSupport) &&
+			(pCM.Name == constants.RootCACertConfigMapName || pCM.Name == constants.TenantRootCACertConfigMapName) {
+			continue
+		}
 		pSet.Insert(differ.ClusterObject{Object: pCM, Key: differ.DefaultClusterObjectKey(pCM, "")})
 	}
 
--- a/virtualcluster/pkg/syncer/resources/configmap/checker_test.go
+++ b/virtualcluster/pkg/syncer/resources/configmap/checker_test.go
@@ -27,11 +27,15 @@ import (
 	core "k8s.io/client-go/testing"
 
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/apis/tenancy/v1alpha1"
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/constants"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/conversion"
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/featuregate"
 	util "sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/test"
 )
 
 func TestConfigMapPatrol(t *testing.T) {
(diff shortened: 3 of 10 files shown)


virtualcluster/pkg/syncer/resources/configmap/checker.go
```diff
@@ -63,6 +64,13 @@ func (c *controller) PatrollerDo() {
 	}
 	pSet := differ.NewDiffSet()
 	for _, pCM := range pConfigMaps {
+		// Ingore both RootCACertConfigMapName and TenantRootCACertConfigMapName from the super.
+		// TenantRootCACertConfigMapName is created from the vRootCACertConfigMap and
+		// RootCACertConfigMapName should be left alone.
+		if featuregate.DefaultFeatureGate.Enabled(featuregate.RootCACertConfigMapSupport) &&
+			pCM.Name == constants.RootCACertConfigMapName || pCM.Name == constants.TenantRootCACertConfigMapName {
```
m-messiah, line 71:
```suggestion
			(pCM.Name == constants.RootCACertConfigMapName || pCM.Name == constants.TenantRootCACertConfigMapName) {
```

virtualcluster/pkg/syncer/resources/configmap/checker_test.go
```diff
@@ -27,11 +27,15 @@ import (
...
 
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/apis/tenancy/v1alpha1"
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/constants"
 	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/conversion"
+	"sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/featuregate"
 	util "sigs.k8s.io/cluster-api-provider-nested/virtualcluster/pkg/syncer/util/test"
 )
 
 func TestConfigMapPatrol(t *testing.T) {
+	featuregate.DefaultFeatureGate.Set(featuregate.RootCACertConfigMapSupport, true)
```
m-messiah, line 37: IIRC, we have SetFeatureGateDuringTest https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/dbcdf7c4b3825340aef3c6bf343fdb0d5cd9e8d0/virtualcluster/pkg/syncer/util/test/featuregate.go#L31-L31

christopherhein: Oh nice!

virtualcluster/pkg/syncer/conversion/equality.go
```diff
@@ -447,6 +448,13 @@ func (e vcEquality) checkInt64Equality(pObj, vObj *int64) (*int64, bool) {
 // CheckConfigMapEquality checks whether super control plane ConfigMap and virtual ConfigMap
 // are logically equal. The source of truth is virtual object.
 func (e vcEquality) CheckConfigMapEquality(pObj, vObj *v1.ConfigMap) *v1.ConfigMap {
+	// Before processing Equality on ConfigMaps we need to reset the CM if pObj is used for
+	// RootCACertConfigMaps.
+	if featuregate.DefaultFeatureGate.Enabled(featuregate.RootCACertConfigMapSupport) &&
```
m-messiah, line 453: I am wondering, can you reuse getConfigMapName func here, as it looks mostly the same?

virtualcluster/pkg/syncer/resources/pod/mutatorplugin/podrootcacertmutator.go
```diff
@@ -0,0 +1,72 @@
...
+type PodRootCACertMutatorPlugin struct{}
+
+// This mutator will automatically reassign configmap references for configmaps named
+// kube-root-ca.crt in the pod spec, these places are
+// * volumes
+// * env
+// * envFrom
+func (pl *PodRootCACertMutatorPlugin) Mutator() conversion.PodMutator {
+	return func(p *conversion.PodMutateCtx) error {
+		if featuregate.DefaultFeatureGate.Enabled(featuregate.RootCACertConfigMapSupport) {
```
m-messiah, line 44: let's flip the if and do fast return if the feature is not enabled? You could even probably add this check before the plugin registration

virtualcluster/pkg/syncer/conversion/equality.go
christopherhein: Yeah, I was considering that, I think if I move that func into this package I can, I don't want to import the configmap package into this conversion cause that would break the flexibility of using vc-syncer as a framework and only enabling _some_ without noticing that others were automatically init'ed. I will move that func here or to a util package.

virtualcluster/pkg/syncer/resources/pod/mutatorplugin/podrootcacertmutator.go
christopherhein: Great idea returning early. 
I'm a little worried about doing that in the init phase incase the feature gates are loaded after, since init is on import… am I wrong about that?

christopherhein: I don't think we can cause we import them from cmd/vc-syncer and in main() from there we also start to load the flags,  I think fast exiting when featuregate is false is enough.

christopherhein:
This is running into a potential test-infra issue on where it's running. https://github.com/kubernetes/test-infra/issues/27919

m-messiah (approved):

virtualcluster/pkg/syncer/resources/configmap/checker.go
```diff
@@ -63,6 +64,13 @@ func (c *controller) PatrollerDo() {
 	}
 	pSet := differ.NewDiffSet()
 	for _, pCM := range pConfigMaps {
+		// Ingore both RootCACertConfigMapName and TenantRootCACertConfigMapName from the super.
```
Fei-Guo, line 67: Question: Should we ignore or compare against the right configMap stored in in vc namespace? There still has a change the configmap in super is altered unexpectedly? (You can make a follow up change if you agree.)

christopherhein: You are right, that is a miss in the current setup. Let me think about how to handle this and I will do a follow up.

christopherhein: Some combination of maybe a label on the `tenant-kube-root-ca.crt` that maps the name back to the `kube-root-ca.crt` and checks isn't contents. I think the one thing we do have going for us is right now when this is enabled we still reconcile the tenants `kube-root-ca.crt` so each time checker runs it pops an update back on the reconciler queue which will try to update the super when there is a difference so I think we actually do cover this but in a semi inefficient way.

wondywang:
Hi @christopherhein, I have added some comments. That may also need to be considered.

virtualcluster/pkg/syncer/resources/pod/mutatorplugin/podrootcacertmutator.go
```diff
@@ -0,0 +1,72 @@
...
+			return nil
+		}
+
+		for i := range p.PPod.Spec.Volumes {
+			if p.PPod.Spec.Volumes[i].ConfigMap != nil && p.PPod.Spec.Volumes[i].ConfigMap.Name == constants.RootCACertConfigMapName {
+				p.PPod.Spec.Volumes[i].ConfigMap.Name = constants.TenantRootCACertConfigMapName
+			}
+		}
+
+		for c := range p.PPod.Spec.Containers {
```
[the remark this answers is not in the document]
christopherhein: Good call! I forgot about initContainers!

virtualcluster/pkg/syncer/resources/configmap/checker.go
```diff
@@ -63,6 +64,13 @@ func (c *controller) PatrollerDo() {
...
+		// TenantRootCACertConfigMapName is created from the vRootCACertConfigMap and
+		// RootCACertConfigMapName should be left alone.
+		if featuregate.DefaultFeatureGate.Enabled(featuregate.RootCACertConfigMapSupport) &&
+			(pCM.Name == constants.RootCACertConfigMapName || pCM.Name == constants.TenantRootCACertConfigMapName) {
+			continue
+		}
```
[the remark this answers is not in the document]
christopherhein: Yea, it's a little wasteful, but that will then requeue it and the DWS Reconcile func will notice it exists and is "equal" even though the names don't align and drop it off the queue.

virtualcluster/pkg/syncer/resources/pod/mutatorplugin/podrootcacertmutator.go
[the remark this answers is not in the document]
christopherhein: https://github.com/kubernetes-sigs/cluster-api-provider-nested/pull/328

PR #330, opened 2022-11-13
✨ Enhancement for virtual cluster DNS

### LINKED ISSUE #329: ✨ Enhancement for virtual cluster DNS
**User Story**

As a user I would like to use different versions of dns server in different tenants.

For example, CoreDNS v1.10.0 in virtual cluster A, and CoreDNS v1.9.2 in virtual cluster B. 

**Detailed Description**
The current DNS solution is using CoreDNS with a specific patch ([tenant-dns doc](https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/doc/tenant-dns.md#solution)). So that we need to rebuild multiple versions of the CoreDNS image, which will take a lot of work.

I would like to introduce a new solution that does not require changes to the code of CoreDNS.
We can leverage the [k8s_external](https://coredns.io/plugins/k8s_external/#examples) plugin in CoreDNS, and then add the cluster ip of `pService` to the `ExternalIPs` of `vService`. So that we can successfully resolve the service name in pod.

/kind feature

LuBingtan:
Enhancement for virtual cluster DNS: https://github.com/kubernetes-sigs/cluster-api-provider-nested/issues/329

Fixes #329

### DIFF
--- a/virtualcluster/pkg/controller/controllers/virtualcluster_controller_test.go
+++ b/virtualcluster/pkg/controller/controllers/virtualcluster_controller_test.go
@@ -179,6 +179,7 @@ var _ = Describe("VirtualCluster Controller", func() {
 			}, time.Minute*5, interval).Should(BeTrue())
 
 			By("Updating ClusterVersion")
+			Expect(cli.Get(ctx, client.ObjectKeyFromObject(cvInstance), cvInstance)).Should(Succeed())
 			cvInstance.Spec.APIServer.StatefulSet.Spec.Template.Spec.Containers[0].Args = append([]string{"-v=7"}, cvInstance.Spec.APIServer.StatefulSet.Spec.Template.Spec.Containers[0].Args...)
 			if cvInstance.Spec.APIServer.Service.Labels == nil {
 				cvInstance.Spec.APIServer.Service.Labels = map[string]string{}
--- a/virtualcluster/config/sampleswithspec/coredns_external.yaml
+++ b/virtualcluster/config/sampleswithspec/coredns_external.yaml
@@ -1,0 +1,196 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: coredns
+  namespace: kube-system
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  labels:
+    kubernetes.io/bootstrapping: rbac-defaults
+  name: system:coredns
+rules:
+- apiGroups:
+  - ""
+  resources:
+  - endpoints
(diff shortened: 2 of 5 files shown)


LuBingtan:
> This looks a little tricky. Is the k8s_external plugin enabled by default? or is there any chance the actual external_ip field will be needed in vc (I know we did not upward synchronization it right now) ?
> 
> The bottom line is that if we want to play the trick in the syncer, we need a feature switch for sure. @christopherhein, will this help in your case?

As far as I know,  the k8s_external plugin is not enabled by default.

As for the question whether the actual external_ip field is needed in vc. According this doc (https://kubernetes.io/docs/concepts/services-networking/service/#external-ips), this field is used for exposing the service on a specific node. That means it is usually set by users on purpose. We can add an `if` and leave this field unchanged in this case.

And I agree with you that we need a feature switch.

christopherhein:
Agree about the feature switch. 

Does this work with DualStacked services too? For us, we're going the route of actually making the services routable to the tenant. This might may be useful for some folks though, the patch is kind of messy.

LuBingtan:
Already add a feature gate. PTAL 😄 

> Does this work with DualStacked services too? For us, we're going the route of actually making the services routable to the tenant. This might may be useful for some folks though, the patch is kind of messy.

As for dual-stack services, I'll give it a try and see if this works.

Fei-Guo:
@LuBingtan Thanks for working on this.

/lgtm

@christopherhein I will leave it to you for approval.

Fei-Guo:
Can you double check the test error?

virtualcluster/pkg/controller/controllers/virtualcluster_controller_test.go
```diff
@@ -179,6 +179,7 @@ var _ = Describe("VirtualCluster Controller", func() {
 			}, time.Minute*5, interval).Should(BeTrue())
 
 			By("Updating ClusterVersion")
+			Expect(cli.Get(ctx, client.ObjectKeyFromObject(cvInstance), cvInstance)).Should(Succeed())
```
LuBingtan, line 182: @Fei-Guo  Could please take a look and trigger the test again?

I found the test case has this error message:
```
Operation cannot be fulfilled on clusterversions.tenancy.x-k8s.io "clusterversion-samplegh6xl": the object has been modified; please apply your changes to the latest version and try again
```
IMHO, the reason for the error is because the `cvInstance` object is staled here. So here I am trying to re-get the object. It works on my local environment.
But this error doesn't seem to have happened in the past, which is a little bit weird.

PR #333, opened 2022-11-23
🐛 Fix port-forward failure

### LINKED ISSUE #332: 🐛[VC] Failed to do port-forward for a pod in virtual cluster
I tried to port-forward for a pod in VC:
- Create a nginx pod in a virtual cluster
- Use a command like this `kubectl port-forward xxxx 9999:80`
- Then I got an error:
```
error: error upgrading connection: pods "xxxx" is forbidden: User "system:serviceaccount:vc-manager:vn-agent" cannot create resource "pods/portForward" in API group "" in the namespace "xxxxx"
```

Port-forward command should be succeeded.

It seems that the `"portForward"` here should be `"portforward"`.
https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/cd4b6eb6e248326e6152f7684ac6d25b01fa5fbf/virtualcluster/pkg/vn-agent/server/translate.go#L123-L125

I tried to fix this and have tested on my local environment, and it was working.

**Environment:**

- cluster-api-provider-nested version: master
- Minikube/KIND version: kind version 0.16.0
- Kubernetes version: (use `kubectl version`): Super cluster: v1.25.2, Virtual cluster: v1.25.0
- OS (e.g. from `/etc/os-release`):

/kind bug

LuBingtan:
Fix #332 
Fixes #332

### DIFF
--- a/virtualcluster/pkg/vn-agent/server/translate.go
+++ b/virtualcluster/pkg/vn-agent/server/translate.go
@@ -121,7 +121,7 @@ func TranslatePathForSuper(req *restful.Request, tenantName string) error {
 		apiserverPath = path.Join(commonPath, "attach")
 		translateRawQuery(req, containerName)
 	case "portForward":
-		apiserverPath = path.Join(commonPath, "portForward")
+		apiserverPath = path.Join(commonPath, "portforward")
 		translateRawQuery(req, "")
 	default:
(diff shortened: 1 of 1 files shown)


wondywang:
I feel it do make sense.

PR #334, opened 2022-12-01
🌱 volume.Projected support for kube-root-ca.crt ConfigMaps

wondywang:
**What this PR does / why we need it** :
In this PR I would like to make volume.Projected support for kube-root-ca.crt ConfigMaps. And meanwhile renew the Checker to check  the TenantRootCACertConfigMapName in super.

**Which issue(s) this PR fixes** :
Fixes #

/kind feature

### DIFF
--- a/virtualcluster/pkg/syncer/resources/configmap/checker_test.go
+++ b/virtualcluster/pkg/syncer/resources/configmap/checker_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2020 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


wondywang:
@christopherhein @Fei-Guo 

PTAL, thanks.

virtualcluster/pkg/syncer/resources/configmap/checker_test.go
```diff
@@ -93,7 +93,9 @@ func TestConfigMapPatrol(t *testing.T) {
 			ExistingObjectInSuper: []runtime.Object{
 				superConfigMap(constants.TenantRootCACertConfigMapName, superDefaultNSName, "12345", defaultClusterKey),
 			},
-			ExpectedNoOperation: true,
+			ExpectedDeletedPObject: []string{
```
Fei-Guo, line 96: It looks like this PR does not cause this behavior change, how can the test pass before?

wondywang: Yes, as I commented  at #326 . The pConfigMap `TenantRootCACertConfigMapName` is ignored by checker.  https://github.com/kubernetes-sigs/cluster-api-provider-nested/blob/main/virtualcluster/pkg/syncer/resources/configmap/checker.go#L71

Here I would let the checker reset the pConfigMap name from `tenant-kube-root-ca.crt` to `kube-root-ca.crt` to compare with the vConfigMap.

Fei-Guo: I understand. thanks for the explanation.

Fei-Guo:
LGTM. 

I will let @christopherhein to approve.

virtualcluster/pkg/syncer/resources/configmap/checker_test.go
wondywang: It can effectively prevent these configmaps from entering the queue. This is beneficial in some extreme cases.

christopherhein (approved):
/lgtm
/approve

Nice work, this is great!

PR #340, opened 2023-02-16
🐛 namespace names should valid RFC 1123 DNS labels

wondywang:
What type of PR is this?
/kind bug

What this PR does / why we need it:
As mentioned here, https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#namespaces-and-dns, all namespace names must be valid RFC 1123 DNS labels.

### DIFF
--- a/virtualcluster/pkg/syncer/conversion/helper.go
+++ b/virtualcluster/pkg/syncer/conversion/helper.go
@@ -60,7 +60,7 @@ func ToClusterKey(vc *v1alpha1.VirtualCluster) string {
 
 func ToSuperClusterNamespace(cluster, ns string) string {
 	targetNamespace := strings.Join([]string{cluster, ns}, "-")
-	if len(targetNamespace) > validation.DNS1123SubdomainMaxLength {
+	if len(targetNamespace) > validation.DNS1123LabelMaxLength {
 		digest := sha256.Sum256([]byte(targetNamespace))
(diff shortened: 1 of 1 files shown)


wondywang:
@christopherhein  PTAL, thanks.

Fei-Guo:
Out of curiosity, what is the difference between these two values?

wondywang:
> Out of curiosity, what is the difference between these two values?

```go
// DNS1123SubdomainMaxLength is a subdomain's max length in DNS (RFC 1123)
const DNS1123SubdomainMaxLength int = 253

// DNS1123LabelMaxLength is a label's max length in DNS (RFC 1123)
const DNS1123LabelMaxLength int = 63
```

Fei-Guo:
Is DNS1123SubdomainMaxLength changed? I clearly remember this should be 64, otherwise the code is wrong.

wondywang:
I'm not sure about this. But I found the latest definition, it is indeed the case. Because I recently discovered that there was too long namespace that was not synchronized, and then I found the reason here.

https://github.com/kubernetes/apimachinery/blob/master/pkg/util/validation/validation.go#L209

Fei-Guo:
Did you change your kubernetes version locally?

wondywang:
> Did you change your kubernetes version locally?

no. and I checked the historical version of k8s.io/apimachinery, it should always be that value.

Fei-Guo:
cool. Thanks for fixing the long-term bug then. 

/lgtm
/approve

PR #345, opened 2023-04-26
✨  Add LeaderElectionResourceLock flag and generalize leaderElectionID flag name

m-messiah:
The PR adds a new flag `--leader-elect-resource-lock` to define the type of leader election resource lock (configmapleases, configmaps or leases). Current default behaviour of controller-runtime sets it to configmapleases, so the controller creates and updates two objects now - ConfigMap and Lease. Adding the flag would help operators to switch controller to use leases only to reduce load to the API and network.

The PR also deprecates the flag `--le-cm-name` because it was using `cm` in it, that could be confusing after the switch to leases; so it introduces the new flag `--leader-elect-resource-name` with the fallback to `--le-cm-name` if it is set explicitly.

### DIFF
--- a/virtualcluster/cmd/manager/main.go
+++ b/virtualcluster/cmd/manager/main.go
@@ -69,12 +69,12 @@ func main() {
 	flag.StringVar(&controlPlaneProvisioner, "provisioner", "native",
 		"The underlying platform that will provision control plane for virtualcluster.")
 	flag.BoolVar(&leaderElection, "leader-election", true, "If enable leaderelection for vc-manager")
-	// Deprecated: the flag used resource type as part of the name. Replaced by le-name.
-	flag.StringVar(&leaderElectionCmName, "le-cm-name", "", "DEPRECATED. Use --le-name instead")
-	flag.StringVar(&leaderElectionID, "le-name", "vc-manager-leaderelection-lock",
+	// Deprecated: the flag used resource type as part of the name. Replaced by leader-elect-resource-name.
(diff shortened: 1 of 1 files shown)


virtualcluster/cmd/manager/main.go
```diff
@@ -67,8 +69,12 @@ func main() {
...
 		"The underlying platform that will provision control plane for virtualcluster.")
 	flag.BoolVar(&leaderElection, "leader-election", true, "If enable leaderelection for vc-manager")
-	flag.StringVar(&leaderElectionCmName, "le-cm-name", "vc-manager-leaderelection-lock",
-		"The name of the configmap that will be used as the resourcelook for leaderelection")
+	// Deprecated: the flag used resource type as part of the name. Replaced by le-name.
+	flag.StringVar(&leaderElectionCmName, "le-cm-name", "", "DEPRECATED. Use --le-name instead")
+	flag.StringVar(&leaderElectionID, "le-name", "vc-manager-leaderelection-lock",
+		"The name of the resource that will be used as the resourcelock for leaderelection")
+	flag.StringVar(&leaderElectionResourceLock, "le-resource", "configmapsleases",
+		"The type of the resource that will be used as the resourcelock for leaderelection [configmaps, leases, configmapsleases]")
```
christopherhein, line 77: Can we change these to line up with core kubernetes leader election flag names - https://github.com/kubernetes/component-base/blob/25833dc857cd1d3dd4ae46014ada3392729729bf/config/options/leaderelectionconfig.go#L25-L53

christopherhein (changes requested):
Thank you for doing this, I’d like to align more with component base flag names if possible. 🙏

christopherhein (approved):
/lgtm
/approve

